PR #1, opened 2017-09-17
Favor using kubectl apply rather than create

ianlewis:
kubectl apply is usually better when creating things from manifests because they can be more easily updated later, commands can be re-run if you hit errors etc.


enisoc:
Sorry for the slow response. I forgot to sign up to watch my own repo. :)

This is a good suggestion. Would you like to resolve the conflicts yourself? If not, I'd be happy to make the change.

enisoc:
I went ahead and updated it. Thanks again!

PR #10, opened 2018-01-22
[WIP] Vitess Operator example

enisoc:
This is still a work in progress, but cluster startup works.

Next steps:
- [ ] Rebase on the new Vitess templates in https://github.com/youtube/vitess/pull/3487.
- [ ] Add back [Orchestrator](https://github.com/github/orchestrator) integration as in the old Vitess Helm chart.
- [ ] Implement Vitess-specific rolling update strategy for tablets: keyspace-by-keyspace, shard-by-shard, reparent before updating master.
- [ ] Automatically initialize replication on new shards.

cc @derekperkins

## Vitess Operator

This is an example of an app-specific [Operator](https://coreos.com/operators/), in this case for [Vitess](http://vitess.io), built with Metacontroller.

It's meant to demonstrate the following patterns:

* Building an Operator for a complex, stateful application out of a set of small Lambda Controllers that each do one thing well.
* Replacing static, client-side template rendering with Lambda Controllers, which can adjust based on dynamic cluster state.
* Using a functional-style language ([Jsonnet](http://jsonnet.org)) to define Lambda Controllers in terms of template-like transformations on JSON objects.
* Using the "Apply" update strategy feature of CompositeController, which emulates the behavior of `kubectl apply`, except that it attempts to do pseudo-strategic merges for CRDs.

See the [README](https://github.com/enisoc/kube-metacontroller/tree/vitess/examples/vitess) for more details.

### DIFF
--- a/examples/vitess/README.md
+++ b/examples/vitess/README.md
@@ -1,0 +1,126 @@
+## Vitess Operator
+
+This is an example of an app-specific [Operator](https://coreos.com/operators/), 
+in this case for [Vitess](http://vitess.io), built with Metacontroller.
+
+It's meant to demonstrate the following patterns:
+
+* Building an Operator for a complex, stateful application out of a set of small
+  Lambda Controllers that each do one thing well.
+  * In addition to presenting a k8s-style API to users, this Operator uses
+    custom k8s API objects to coordinate within itself.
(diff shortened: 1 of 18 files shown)


derekperkins:
This looks great! Time to learn jsonnet :)

PR #11, opened 2018-01-23
Add client code generators

### LINKED ISSUE #3: Generate a client and informer for Metacontroller APIs
We need to use the dynamic client for the objects that Lambda Controllers manage, since we allow them to manage arbitrary objects.

However, there are advantages to using a generated client to access objects that represent the Metacontroller APIs themselves (e.g. `CompositeController`).

[This blog](https://blog.openshift.com/kubernetes-deep-dive-code-generation-customresources/) may be useful.

antoineco:
Generate and use client code using Kubernetes' code generators.

Closes #3


antoineco:
Note: the "use" part is still missing, submitting PR for initial review of code generation.

enisoc:
Looks like a great start!

antoineco:
@enisoc I'm ready for a review.
I split the usage of clientset and informers(+listers) into 2 different commit since the first one doesn't strictly require the second.

main.go
```diff
@@ -19,18 +19,49 @@ package main
...
+	internallisters "k8s.io/metacontroller/client/generated/lister/metacontroller/v1alpha1"
 
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
 )
 
-func resyncAll(config *rest.Config) error {
+// metaController manages metacontroller API objects.
+type metaController struct {
```
enisoc: This seems to hold the same place semantically as our `SharedInformerFactory`, just with aliases for some long "getter" calls. Were you thinking there would be additional logic here? If not, I'd prefer to just pass around the `SharedInformerFactory` directly, so we don't have to keep this struct in sync with it.

main.go
```diff
@@ -64,12 +89,45 @@ func resyncAll(config *rest.Config) error {
...
 
+	// Create OS signal channels.
+	sigs := make(chan os.Signal, 1)
+	stop := make(chan struct{})
+	signal.Notify(sigs, os.Interrupt, syscall.SIGTERM)
+
+	go func() {
+		<-sigs
+		glog.V(2).Info("Stopping informers")
+		close(stop)
```
enisoc: Unfortunately, SharedInformerFactory doesn't give us any way to wait for the informers to cleanly stop. Since we're about to terminate the process anyway, I don't think we gain anything by closing this channel. I suspect stopping was only implemented to avoid unbounded resource growth during tests that start and stop many informers within one process.

With that said, it doesn't hurt to close it. We can also use the same channel to stop our infinite poll loop, which would let us avoid `os.Exit()` as mentioned below.

main.go
```diff
@@ -64,12 +89,45 @@ func resyncAll(config *rest.Config) error {
...
+		os.Exit(0)
```
enisoc: I try to avoid `os.Exit(0)` whenever possible because it skips deferred functions. For handling SIGTERM, I prefer to return a 0 exit code by falling off the end of `main()`, which follows the Go idiom of the happy path flowing all the way through the function:

https://github.com/GoogleCloudPlatform/kube-metacontroller/blob/94c57bb47f286164b967b983421131cac2faad2e/examples/jsonnetd/main.go#L88-L93

In this case, that would mean moving the infinite poll loop into a goroutine, and breaking when the `stop` channel is ready. We'd also want a `done` channel that gets closed when the polling goroutine returns.

What do you think? I could also restructure this part after merging if you prefer.

composite_controller.go
```diff
@@ -35,14 +34,14 @@ import (
...
-	ccList, err := metaClient.MetacontrollerV1alpha1().CompositeControllers("").List(metav1.ListOptions{})
+func syncAllCompositeControllers(dynClient *dynamicClientset, mc *metaController) error {
+	ccList, err := (mc.ccLister).List(labels.NewSelector())
 	if err != nil {
 		return fmt.Errorf("can't list CompositeControllers: %v", err)
 	}
 
-	for i := range ccList.Items {
-		cc := &ccList.Items[i]
+	for i := range ccList {
```
enisoc: Now that we're already getting a slice of pointers, we can simplify this to:

```go
for _, cc := range ccList {
```

Same for `icList` below.

composite_controller.go
```diff
@@ -35,14 +34,14 @@ import (
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 )
 
-func syncAllCompositeControllers(dynClient *dynamicClientset, metaClient *internalclient.Clientset) error {
-	ccList, err := metaClient.MetacontrollerV1alpha1().CompositeControllers("").List(metav1.ListOptions{})
+func syncAllCompositeControllers(dynClient *dynamicClientset, mc *metaController) error {
+	ccList, err := (mc.ccLister).List(labels.NewSelector())
```
enisoc: We've seen bugs before caused by the ambiguity of whether the zero-value of `Selector` means "select everything" or "select nothing". Let' use `labels.Everything()` here so it's clear what we mean. Same for `icList` below.

main.go
antoineco: That's reasonable. I usually use this pattern where structs contain other references like workqueues, "additional logic" to quote your words. No real need for this here.

main.go
antoineco: That makes sense. To be honest I added the whole signal handling part just to be able to [throw *something* at the stop channel](https://github.com/kubernetes/client-go/blob/ad1ee6b/informers/factory.go#L81). I'm aware doing this without even waiting defies the purpose.

I can either pass a dummy channel to the `sharedInformerFactory` and remove this entire routine, or duplicate what you did in the jsonnetd example. What are you leaning towards to?

enisoc: Ah, in that case let's just pass `nil` to the stop channel for now, to keep this PR solely about the generated client/informer. We can add graceful signal handling in a separate PR. Let me know if you plan to do that, or else I'll whip something up.

antoineco: I'll pass `nil` then. Feel free to deal with the signals, you seem to have a clear idea of how you want to implement that stop -> done -> exit logic.

enisoc:
@antoineco It looks like everything's been addressed. Is this ready to merge?

antoineco:
From my side yes.

PR #12, opened 2018-01-25
Add dockerignore file

antoineco:
Ignore everything except the built artifact.

No big deal but sending large `vendor/` and `.git/` directories is quite unnecessary.

Before/after
```
Sending build context to Docker daemon  117.1MB
---
Sending build context to Docker daemon  29.24MB
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+*
+!metacontroller


enisoc:
Good idea. Thanks!

PR #34, opened 2018-03-10
examples/go: Add skeleton for Go hooks

enisoc:
This controller doesn't do anything useful. It's just an example skeleton for writing Metacontroller hooks with Go.


examples/go/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM golang:1.10 AS build
```
luisdavim: Can you use `golang:1.11`?

examples/go/thing-controller.yaml
```diff
@@ -0,0 +1,61 @@
...
+    resource: things
+  childResources:
+    - apiVersion: v1
+      resource: pods
+  hooks:
+    sync:
+      webhook:
+        url: http://thing-controller.metacontroller/sync
+---
+apiVersion: apps/v1beta1
```
luisdavim: Can you use `apps/v1`?

enisoc:
I posted this as an example, but I don't think it's ready to merge. Given the problems with generating JSON from statically-typed languages (#76), I'd like to limit the examples to dynamic languages for now.

luisdavim:
I understand but you could just write that in a disclaimer in the readme and make it available for people that want to use go.

enisoc:
Alright that's a good point. I'll clean it up and add the disclaimer as you suggested.

PR #36, opened 2018-03-15
pint to 1.9.1 k8s release dependency

etsangsplk:
to avoid flux in api changes in k8s, pin to release versions of k8s deps.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -15,9 +15,12 @@
 
 [[projects]]
   name = "github.com/gogo/protobuf"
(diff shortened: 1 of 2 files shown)


enisoc:
Thanks for this! I'm currently working on getting the repo moved to a new home, but I'll come back and merge this once that settles.

etsangsplk:
@enisoc Cool. A few of my coworkers basically says they cannot persuade their teams to use it if the dependencies are not locked down to releases.

etsangsplk:
I signed it!

PR #50, opened 2018-05-25
Add flag to set custom CA

bakins:
I generally run all my services inside the cluster with certificates signed by an internal CA. I want to run my metacontroller hooks the same way.  This PR adds a flag to allow adding a custom CA for certificate verification.

I wound up having to touch a lot more code that I liked. Ultimately, I needed to use a custom `http.Client` in `hooks/webhook.go`, but I need to pass that through from `main.go`. I'm not sure this is the best approach, but wanted to offer it up for discussion.

Thanks for the project! This allowed me to greatly simplify some custom controllers I maintain at my day job.

### DIFF
--- a/main.go
+++ b/main.go
@@ -17,7 +17,11 @@ limitations under the License.
 package main
 
 import (
+	"crypto/tls"
+	"crypto/x509"
 	"flag"
+	"io/ioutil"
+	"net/http"
 	"os"
 	"os/signal"
 	"sync"
@@ -42,6 +46,7 @@ import (
 var (
 	discoveryInterval = flag.Duration("discovery-interval", 30*time.Second, "How often to refresh discovery cache to pick up newly-installed resources")
 	informerRelist    = flag.Duration("cache-flush-interval", 30*time.Minute, "How often to flush local caches and relist objects from the API server")
+	caFile            = flag.String("ca", "", "Verify certificates of TLS-enabled secure servers using this CA bundle")
 )
 
 type controller interface {
@@ -73,6 +78,8 @@ func main() {
 	}
 	mcInformerFactory := mcinformers.NewSharedInformerFactory(mcClient, *informerRelist)
 
+	httpClient := createHTTPClient(*caFile)
+
 	// Create dynamic clientset (factory for dynamic clients).
 	dynClient := dynamicclientset.New(config, resources)
 	// Create dynamic informer factory (for sharing dynamic informers).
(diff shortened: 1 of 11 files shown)


main.go
```diff
@@ -42,6 +46,7 @@ import (
 var (
 	discoveryInterval = flag.Duration("discovery-interval", 30*time.Second, "How often to refresh discovery cache to pick up newly-installed resources")
 	informerRelist    = flag.Duration("cache-flush-interval", 30*time.Minute, "How often to flush local caches and relist objects from the API server")
+	caFile            = flag.String("ca", "", "Verify certificates of TLS-enabled secure servers using this CA bundle")
```
enisoc: Can we call the flag `ca-bundle-file` so it's clear what it does when looking at a command line in isolation?

main.go
```diff
@@ -110,3 +117,32 @@ func main() {
...
+// Create HTTP client, optionally adding a custom CA
+func createHTTPClient(ca string) *http.Client {
+	httpClient := &http.Client{Timeout: 10 * time.Second}
+	if ca != "" {
+		data, err := ioutil.ReadFile(ca)
+		if err != nil {
+			glog.Fatalf("Can't read CA file %s: %v", ca, err)
+		}
+
+		// an error is okay here, we willl only use the custom CA
```
enisoc: s/willl/will/

main.go
```diff
@@ -110,3 +117,32 @@ func main() {
...
+		rootCAs, _ := x509.SystemCertPool()
```
enisoc: I'd prefer to save the `err` and check both values. In general, we shouldn't make assumptions about the meaning of other returned values if `err` is ever non-nil.

```go
rootCAs, err := x509.SystemCertPool()
if err != nil || rootCAs == nil {
  rootCAs = x509.NewCertPool()
}
```

bakins:
@enisoc Thanks for the feedback. I've been traveling, but will address these soon.

bsherrod:
Looking forward to this change -- we will def use this.  Some basic docs on how to set up would be helpful too.

enisoc:
@bsherrod I think the original author of this PR wasn't able to finish it. I left it open as bait to see if anyone else wanted the same thing. Would you be interested in helping to clone this and get it finished? I'm focusing on integration tests for now and may not get to this for a while otherwise.

bsherrod:
I'll see if someone on my team can take it.

PR #55, opened 2018-06-08
Add user-defined webhook timeout

### LINKED ISSUE #54: [Feature Request] - User configurable webhook timeout.
I recently discovered this project and it looks really great!

### Context
I was considering an application for which `metacontroler` would fit perfectly. A concern I have is that as part of my controller, I would need to make a couple REST calls to an external API to resolve certain conditions at run time. I learned from the [source code](https://github.com/GoogleCloudPlatform/metacontroller/blob/d2bcf42f89860448e63a089720b959be4b4686da/hooks/webhook.go#L34) that the `hook` calls have a hard-coded deadline of 10 seconds, which is quite limiting.

Is there a reason for this particular value (10s)? I assume it shouldn't be a big issue to rise this limit or make it user-defined if the hook calls are being called asynchronously under the hood (and assuming the hook handler is replicated). 

Additional question: How would `metacontroler` handle error conditions on the hook (e.g. return status `!200` or timeouts)? Will the call be skipped, retried, etc?

### Request:

Add an option for the webhook configuration similar to:

```yaml
spec:
  hooks:
    sync:
      webhook:
        url: http://hello-controller.hello/sync
        timeoutSeconds: 60               # Timeout in seconds for the webhook 
```

xocoatzin:
Fixes #54

### Adds: 
- User defined timeout for sync webhooks, defaults to current behavior.
- Documentation.

_Note that I have not tested this PR as I'm still figuring out how to setup the build environment properly._

### DIFF
--- a/apis/metacontroller/v1alpha1/types.go
+++ b/apis/metacontroller/v1alpha1/types.go
@@ -108,7 +108,7 @@ type Hook struct {
 
 type Webhook struct {
 	URL *string `json:"url,omitempty"`
-	TimeoutSeconds *int32 `json:"timeoutSeconds,omitempty"`
+	Timeout *metav1.Duration `json:"timeout,omitempty"`
 
 	Path    *string           `json:"path,omitempty"`
 	Service *ServiceReference `json:"service,omitempty"`
--- a/apis/metacontroller/v1alpha1/types.go
+++ b/apis/metacontroller/v1alpha1/types.go
@@ -108,6 +108,7 @@ type Hook struct {
 
 type Webhook struct {
 	URL *string `json:"url,omitempty"`
+	TimeoutSeconds *int32 `json:"timeoutSeconds,omitempty"`
 
 	Path    *string           `json:"path,omitempty"`
 	Service *ServiceReference `json:"service,omitempty"`
--- a/docs/_api/compositecontroller.md
+++ b/docs/_api/compositecontroller.md
@@ -371,5 +371,5 @@ sent to [`kubectl apply`][kubectl apply].
 That is, you should [set only the fields that you care about](/api/apply/).
 
 Note that your webhook handler must return a response with a status code of `200`
-to be consider successful. Metacontroller will wait for a response for up to the
-amount defined in `timeout`, which defaults to `10s` if not defined explicitly.
+to be considered successful. Metacontroller will wait for a response for up to the
+amount defined in the [Webhook spec](/api/hook/#webhook).
--- a/docs/_api/compositecontroller.md
+++ b/docs/_api/compositecontroller.md
@@ -47,6 +47,7 @@ spec:
     sync:
       webhook:
         url: http://catset-controller.metacontroller/sync
+        timeoutSeconds: 10
 ```
 
 ## Spec
@@ -368,3 +369,7 @@ response because they're in different forms.
 Instead, you should think of each entry in the list of `children` as being
 sent to [`kubectl apply`][kubectl apply].
 That is, you should [set only the fields that you care about](/api/apply/).
+
+Note that your webhook handler must return a response with a status code of `200`
+to be consider successful. Metacontroller will wait for a response for up to the
+amount defined in `timeoutSeconds`, which defaults to `10` if not defined explicitly.
--- a/docs/_api/compositecontroller.md
+++ b/docs/_api/compositecontroller.md
@@ -47,7 +47,7 @@ spec:
     sync:
       webhook:
         url: http://catset-controller.metacontroller/sync
-        timeoutSeconds: 10
+        timeout: 10s
 ```
 
 ## Spec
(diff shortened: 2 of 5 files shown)


hooks/webhook.go
```diff
@@ -101,3 +98,17 @@ func webhookURL(webhook *v1alpha1.Webhook) (string, error) {
 	}
 	return fmt.Sprintf("%s://%s.%s:%v%s", protocol, webhook.Service.Name, webhook.Service.Namespace, port, *webhook.Path), nil
 }
+
+func webhookTimeout(webhook *v1alpha1.Webhook) (time.Duration, error) {
+	if webhook.TimeoutSeconds != nil {
```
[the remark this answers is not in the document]
xocoatzin: Fixed ... *facepalm*

apis/metacontroller/v1alpha1/types.go
```diff
@@ -108,6 +108,7 @@ type Hook struct {
 
 type Webhook struct {
 	URL *string `json:"url,omitempty"`
+	TimeoutSeconds *int32 `json:"timeoutSeconds,omitempty"`
```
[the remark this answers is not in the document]
xocoatzin: Changed. Like it better now, thanks for the feedback.

xocoatzin:
I was able to test this PR on `minikube` and seems to work as expected.

docs/_api/compositecontroller.md
```diff
@@ -368,3 +369,7 @@ response because they're in different forms.
 Instead, you should think of each entry in the list of `children` as being
 sent to [`kubectl apply`][kubectl apply].
 That is, you should [set only the fields that you care about](/api/apply/).
+
+Note that your webhook handler must return a response with a status code of `200`
+to be consider successful. Metacontroller will wait for a response for up to the
```
enisoc: typo: "considered" (same in the other file)

docs/_api/compositecontroller.md
```diff
@@ -368,3 +369,7 @@ response because they're in different forms.
...
+amount defined in `timeout`, which defaults to `10s` if not defined explicitly.
```
enisoc: Since this is under the "Sync Hook Response" heading, readers might be confused where exactly the `timeout` field lives. To make it clear, and to consolidate the info in only one place, I suggest linking to the general Webhook docs like:

```md
... for up to the timeout defined in the [Webhook spec](/api/hook/#webhook).
```

The reader can click through to find the default value, so we don't have to update it everywhere if it changes.

docs/_api/hook.md
```diff
@@ -25,6 +25,7 @@ Each Webhook has the following fields:
 | Field | Description |
 | ----- | ----------- |
 | url | A full URL for the webhook (e.g. `http://my-controller-svc/hook`). If present, this overrides any values provided for `path` and `service`. |
+| timeout | A duration indicating the time that Metacontroller should wait for a response. If the webhook takes longer than this time, the webhook call is aborted and retried later. Defaults to 10s. |
```
enisoc: Let's add a note on the format:

```md
A duration (in the format of Go's time.Duration) ...
```

hooks/webhook.go
```diff
@@ -101,3 +99,19 @@ func webhookURL(webhook *v1alpha1.Webhook) (string, error) {
 	}
 	return fmt.Sprintf("%s://%s.%s:%v%s", protocol, webhook.Service.Name, webhook.Service.Namespace, port, *webhook.Path), nil
 }
+
+func webhookTimeout(webhook *v1alpha1.Webhook) (time.Duration, error) {
+	zero := time.Duration(0)
```
enisoc: I think you can just compare directly against a literal `0` below. Is there some reason to prefer a named variable?

hooks/webhook.go
```diff
@@ -101,3 +99,19 @@ func webhookURL(webhook *v1alpha1.Webhook) (string, error) {
...
+
+	if webhook.Timeout == nil {
+		// Defaults to 10 Seconds to preserve current behavior.
+		return 10 * time.Second, nil
+	}
+
+	if (*webhook.Timeout).Duration <= zero {
```
enisoc: I think it should work as just `webhook.Timeout.Duration` since `.` dereferences as needed. Same below.

docs/_api/compositecontroller.md
xocoatzin: Fixed

docs/_api/compositecontroller.md
xocoatzin: Fixed

docs/_api/hook.md
xocoatzin: Added

hooks/webhook.go
xocoatzin: Changed

> Is there some reason to prefer a named variable?

My complete lack of understanding of go / k8s conventions. I was using [this](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/componentconfig/v1alpha1/defaults.go#L47) as a reference.

hooks/webhook.go
xocoatzin: Changed

PR #59, opened 2018-06-15
MapController Design

enisoc:
This is a proposal for a new Metacontroller API that attempts to fill a gap in the set of supported controller patterns that has been identified based on user feedback. Namely, the proposed API offers a way to observe objects you don't own and tie them into your custom behavior without trying to adopt (take ownership of) those objects.


docs/_design/map-controller.md
```diff
@@ -0,0 +1,399 @@
...
+    output_list.append(map_hook(input_object))
+
+  reconcile_objects(output_list)
+```
+
+where `map_hook()` is the only code that the MapController user writes,
+as a lambda hook.
+
+In general, MapController addresses use cases that can be described as,
+"For every matching X that already exists, I want to create a Y according to the
```
crimsonfaith91: To confirm, we can create multiple Ys right?
And in this case, does the parent object = X = non-owned object, and Y = child object?

enisoc: Good point. I've updated the wording to clarify there can be any number of Y objects for each X. Thanks!

docs/_design/map-controller.md
```diff
@@ -0,0 +1,399 @@
...
+as spitting out a few different things.
+We shouldn't force them to create a CompositeController too just so they can
+emit multiple outputs, especially if those outputs are not conceptually part of
+one larger whole.
+
+For a given input object, the user can generate any number of output objects.
+We will [tag](#map-key) those output objects in some way to associate them with
+the object that we sent as input.
+The tag makes it possible to group those objects and send them along with future
+[map hook requests](#map-hook-request).
```
janetkuo: I'm not sure if I understand the MxN input/output case. IIUC, we tag each output project on which input object they'll be generated from?

docs/_design/map-controller.md
```diff
@@ -0,0 +1,399 @@
...
+
+If the associated input object is gone, we will detect that there is a group of
+output objects that have become orphaned.
```
janetkuo: Not sure if I understand this part. Who orphaned the output objects?

docs/_design/map-controller.md
```diff
@@ -0,0 +1,399 @@
...
+
+The input resources (in this case just PVC) are the inputs to the conceptual
+"map" function.
+We allow multiple input resources because users might want to write a controller
+that performs the same action for several different input types.
+We shouldn't force them to create multiple MapControllers with largely identical
+behavior.
+
+The duck-typed `spec.selector` field (assumed to be `metav1.LabelSelector`) in
+the parent object is used to filter which input objects to process.
```
janetkuo: Why not put `selector` in MapController's `spec.parentResource`?

docs/_design/map-controller.md
```diff
@@ -0,0 +1,399 @@
...
+    resource: volumesnapshots
+  resyncPeriodSeconds: 5
+  hooks:
+    map:
+      webhook:
+        url: http://snapshotschedule-controller.metacontroller/map
+    tombstone:
+      webhook:
+        url: http://snapshotschedule-controller.metacontroller/tombstone
+```
```
janetkuo: IIUC, parent owns outputs (label/selector + ownerRef), and selects inputs (label/selecto only)?

Can a user make map controller enter crazy creation loop by having the same resource in input and output?

docs/_design/map-controller.md
janetkuo: Nvmd you need selector in the parent resource spec

docs/_design/map-controller.md
```diff
@@ -0,0 +1,399 @@
...
+1. It's unclear whether decorator objects are expected to take part in ownership
+   of the objects created.
+   Depending on the use case, users might want attachments to be owned by just
+   the parent, just the decorator, or both.
+   This configuration adds to the cognitive overhead of using the API,
+   and there's no one default that's more intuitive than the others.
+
+## Example
+
+The example use case we'll consider in this doc is a controller called
```
crimsonfaith91: I think this has the potential to be integrated to stateful operators that we are working on. :)

docs/_design/map-controller.md
```diff
@@ -0,0 +1,399 @@
...
+that the user intends to create and hold in the conceptual "bucket" that the
+parent object represents.
+We allow multiple output resources because users might think of their controller
+as spitting out a few different things.
+We shouldn't force them to create a CompositeController too just so they can
+emit multiple outputs, especially if those outputs are not conceptually part of
+one larger whole.
+
+For a given input object, the user can generate any number of output objects.
+We will [tag](#map-key) those output objects in some way to associate them with
```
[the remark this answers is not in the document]
enisoc: Oops! Removed the link for now. I was originally going to write a section about how to actually do the tagging (label vs annotation? name vs uid? etc.), but then I decided that's more of an implementation detail that isn't needed to decide whether MapController is the right abstraction for the set of problems we're targeting.

docs/_design/map-controller.md
enisoc: On the input side, the fact that there can be multiple input resources is only used for selecting the pool of input objects. In pseudocode, something like this:

```go
inputObjects := []Object{}
for _, inputResource := range inputResources {
  inputObjects = append(inputObjects, getMatchingObjects(inputResource, parentSelector)...)
}
```

Then we go through each of the input objects and call the user's hook to generate the output objects for that one input. We tag each output with the input that it came from, which also allows us to select the observed outputs belonging to this input. Continuing in pseudocode:

```go
outputObjects := []Object{}
for _, inputObject := range inputObjects {
  // Compute some opaque string identifying this input object.
  mapKey := makeMapKey(inputObject)

  // Gather observed objects of the output resources that are tagged with this key.
  observedOutputs := []Object{}
  for _, outputResource := range outputResources {
    // Gather all outputs owned by this parent.
    allOutputs := getOwnedObjects(outputResource, parent)
    // Filter to only those tagged for this input.
    observedOutputs = append(observedOutputs, filterByMapKey(allOutputs, mapKey)...)
  }

  // Call user's map hook, passing observed state.
  mapResult := mapHook(parent, inputObject, observedOutputs)
  for _, obj := range mapResult.Outputs {
    // Tag outputs to identify which input they came from.
    setMapKey(obj, mapKey)
    outputObjects = append(outputObjects, obj)
  }
}
reconcileObjects(outputObjects)
```

Does that help? If so, I can add this to the doc somewhere.

docs/_design/map-controller.md
enisoc: I added some clarification that this doesn't refer to orphaning in the ownerRef/controllerRef sense. I wish there was another word I could use that doesn't have a collision with another common API machinery term, but I can't think of one yet. Any ideas?

crimsonfaith91: Will `disconnected` be a good word? Perhaps can set up a new notion of `connection`, which is different from `orphaning` and `adoption`.

docs/_design/map-controller.md
```diff
@@ -0,0 +1,404 @@
...
+        url: http://snapshotschedule-controller.metacontroller/map
+    tombstone:
+      webhook:
+        url: http://snapshotschedule-controller.metacontroller/tombstone
+```
+
+### Parent Resource
+
+The parent resource is the SnapshotSchedule itself, and anything this controller
+creates will by owned by this parent.
```
crimsonfaith91: nit: `will be owned`

docs/_design/map-controller.md
crimsonfaith91: Is there any example of use cases involving same type of input and output resources? I think it will be a very infrequent use case for `MapController`. IIUC, both input and output resources will be different even when they are of same type (because the output resources are new resources created by the `MapController`). Hence, I think this use case may be better served by other meta-controllers.

docs/_design/map-controller.md
```diff
@@ -0,0 +1,404 @@
...
+| `input` | The input object, like what you might get from `kubectl get <input-resource> <input-name> -o json`. |
+| `outputs` | An associative array of output objects that the parent already created for the given input object. |
+
+#### Map Hook Response
+
+| Field | Description |
+| ----- | ----------- |
+| `outputs` | A list of JSON objects representing all the desired outputs for the given input object. |
+
+### Tombstone Hook
```
crimsonfaith91: I think tombstone hook is a more flexible design (e.g., user wants to retain type A output object, but not other types of output objects), but wish to verify whether its role can be technically achieved by existing propagation policy and finalizer feature of garbage collection.

docs/_design/map-controller.md
crimsonfaith91: Another suggestion is the notion of `decay`, which may have a more vivid tone. If an input object is removed, the output objects that are not deleted are said to be `decaying`.

docs/_design/map-controller.md
crimsonfaith91: As synced by person, there are some use cases involving same type of input and output resources like:
1. pod as input, another pod as output
2. service as input, companion service as output
3. ConfigMap with data A as input, ConfigMap with data B as output

crimsonfaith91:
/lgtm
The design is very nicely done. :)

docs/_design/map-controller.md
janetkuo: > Since we put ownerRef on the outputs like you mentioned, we could ignore our own outputs when looking at inputs. 

This stops resources from using their own children as inputs, which seems fine. 

> We'll never prevent all possible hot loops (you can set up multiple controllers, for example) so I just see our responsibility as providing reasonable guardrails against common mistakes.

Agree. Adding a note/caveat is probably enough.

docs/_design/map-controller.md
janetkuo: I see. I did confuse "orphaning" with ownerRef orphaning. Because parent/child metaphor is used for parent & outputs, the use of the term "orphan" could cause confusion. The clarification helps. 

Proposing some more alternative terms: "lost", "untraceable".

docs/_design/map-controller.md
janetkuo: Yes, this helps.

janetkuo:
I replied to some of the comments above.

docs/_design/map-controller.md
enisoc: Do you mean, could we use finalizers as an alternative to the tombstone hook? If so, I think the problem is that the objects would go into a "pending deletion" state, which cannot be undone if the controller later decides to keep the object (e.g. a matching input PVC gets recreated, so the SnapshotSchedule controller decides to keep the "detached" VolumeSnapshot output objects after all).

crimsonfaith91: Noted. Tombstone hook is necessary because "pending deletion" state cannot be undone.

PR #64, opened 2018-07-09
Pin code generator versions

enisoc:
The old way of always fetching the latest generators breaks CI.

The first commit is my changes. The second commit is just the result of running `make generated_files` after updating to the latest v1.9.x code generator versions.


liyinan926:
LGTM. I noticed that code generation was not performed as part of the build process in Dockerfile. So it always uses the checked-in generated client/informer/lister?

enisoc:
Yes, the build just takes the checked-in code as is. When generated code is checked in, running generators again as part of CI or the build process just creates confusion in my experience. If there's ever a difference between the checked-in code and what would be generated, I'd rather build what's checked in (and check in a fix if necessary) than silently build something different than what's checked in.

liyinan926:
Makes sense.

crimsonfaith91:
/lgtm
after merging...
Thanks for tagging me!

rlguarino:
If anything Ive found that having a lint that warns the checked in code is different that what would be generated is sometimes helpful.

PR #65, opened 2018-07-11
README: add note to vendor as k8s.io/metacontroller

nikhita:
Because of `k8s.io/metacontroller` is used as the import path [everywhere](https://github.com/GoogleCloudPlatform/metacontroller/search?q=%22k8s.io%2Fmetacontroller%22&unscoped_q=%22k8s.io%2Fmetacontroller%22).

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,8 @@ the add-on works the same in any Kubernetes cluster.
 
 Please see the [documentation site](https://metacontroller.app) for details.
 
+Note: Vendor this package as `k8s.io/metacontroller`.
+
 ## Contact
 
 Please file [GitHub issues](issues) for bugs, feature requests, and proposals.


README.md
```diff
@@ -9,6 +9,8 @@ the add-on works the same in any Kubernetes cluster.
 
 Please see the [documentation site](https://metacontroller.app) for details.
 
+Note: Vendor this package as `k8s.io/metacontroller`.
```
nikhita: IMO this looks a little out of place but I am not sure where this should really go.

rlguarino: Isnt there an import comment that tells automated toola how to vebdor/download package with a vanity url?

enisoc:
I used to have a similar note in the README, but I moved everything to the docs site and tried to keep README bare so there's only one source of truth:

https://metacontroller.app/contrib/build/

I guess the case you've brought up is a special one, where someone wants to vendor the code, but isn't actually using Metacontroller, so they wouldn't think to look at the user docs site. This repo was never really intended to be imported as a dependency though. In fact, maybe we should move these packages into an `internal/` dir to make it clear that we aren't promising to maintain compatibility for external consumers.

@nikhita Is your end goal to have some other repo depend on packages from this one? If so, maybe we should upstream packages that make sense to be shared to [controller-runtime](https://github.com/kubernetes-sigs/controller-runtime). There might already be viable alternatives implemented there for some things I had to build here. cc @pwittrock

@rlguarino The import path comment sounds like a good idea. I can't use `k8s.io/metacontroller` though because that's not a working URL. I did set up `metacontroller.app` to work as a potential vanity import path, but I didn't rewrite the imports yet because I'm still hoping to find a more vendor-neutral home for this repo, and I don't want to have to change everything a second time to match the conventions of such a home. For example, we should also change the API group (`metacontroller.k8s.io`) if we are moving away from `k8s.io/metacontroller`.

nikhita:
> I guess the case you've brought up is a special one, where someone wants to vendor the code, but isn't actually using Metacontroller, so they wouldn't think to look at the user docs site.

guilty as charged. :)

> @nikhita Is your end goal to have some other repo depend on packages from this one? 

No, I was working on porting the dynamic informers/listers from this repo to upstream (https://github.com/kubernetes/kubernetes/issues/64319#issuecomment-401365610). Since I wasn't really "using" this repo (just reading and understanding the source code initially), I git cloned it as `GoogleCloudPlatform/metacontroller` in my GOPATH, which led to several errors as all the import paths were `k8s.io/metacontroller`.

I'll close this PR because there is a neat docs website detailing everything. I don't think someone will really do what I was trying to do :) and anyone who uses it will surely read the docs. 

> I can't use k8s.io/metacontroller though because that's not a working URL. I did set up metacontroller.app to work as a potential vanity import path, but I didn't rewrite the imports yet because I'm still hoping to find a more vendor-neutral home for this repo, and I don't want to have to change everything a second time to match the conventions of such a home. 

Makes sense! :+1: 

Thanks @rlguarino @enisoc!

PR #66, opened 2018-07-17
only allow one Metacontroller instance to be active at a time

crimsonfaith91:
Addressing issue #5 partially. In the long run, we still have to resolve leader election issue when the user spawns more than one Metacontroller instance.

Metacontroller uses `apps/v1beta1 Deployment` under the hood. This PR changes the Metacontroller to use `apps/v1beta2 StatefulSet` to maintain the invariant of only one Metacontroller instance is active at a time.

### DIFF
--- a/manifests/dev/args.yaml
+++ b/manifests/dev/args.yaml
@@ -1,6 +1,6 @@
 # Override args for development mode.
-apiVersion: apps/v1beta1
-kind: Deployment
+apiVersion: apps/v1
+kind: StatefulSet
 metadata:
   name: metacontroller
   namespace: metacontroller
--- a/manifests/dev/args.yaml
+++ b/manifests/dev/args.yaml
@@ -5,6 +5,7 @@ metadata:
   name: metacontroller
   namespace: metacontroller
 spec:
+  serviceName: ""
   template:
     spec:
       containers:
--- a/manifests/dev/args.yaml
+++ b/manifests/dev/args.yaml
@@ -1,5 +1,5 @@
 # Override args for development mode.
-apiVersion: apps/v1
+apiVersion: apps/v1beta2
 kind: StatefulSet
 metadata:
   name: metacontroller
--- a/manifests/dev/args.yaml
+++ b/manifests/dev/args.yaml
@@ -5,7 +5,6 @@ metadata:
   name: metacontroller
   namespace: metacontroller
 spec:
-  serviceName: ""
   template:
     spec:
       containers:
--- a/manifests/metacontroller.yaml
+++ b/manifests/metacontroller.yaml
@@ -43,7 +43,7 @@ spec:
     singular: controllerrevision
     kind: ControllerRevision
 ---
-apiVersion: apps/v1
+apiVersion: apps/v1beta2
 kind: StatefulSet
 metadata:
   name: metacontroller
--- a/manifests/metacontroller.yaml
+++ b/manifests/metacontroller.yaml
@@ -55,6 +55,7 @@ spec:
   selector:
     matchLabels:
       app: metacontroller
+  serviceName: ""
   template:
     metadata:
       labels:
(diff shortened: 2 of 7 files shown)


crimsonfaith91:
@enisoc Do we want to use `StatefulSet` for hooks too? For example, do we need to change [this](https://github.com/GoogleCloudPlatform/metacontroller/blob/master/examples/bluegreen/bluegreen-controller.yaml#L39)?

manifests/metacontroller.yaml
```diff
@@ -43,8 +43,8 @@ spec:
     singular: controllerrevision
     kind: ControllerRevision
 ---
-apiVersion: apps/v1beta1
-kind: Deployment
+apiVersion: apps/v1
```
enisoc: Let's use `apps/v1beta2` for now so we maintain compatibility with k8s 1.8.

crimsonfaith91: noted!

enisoc:
> Do we want to use StatefulSet for hooks too?

For hooks, we don't need to guarantee that there isn't more than one, but we do ideally want to ensure that all replicas are the same version at a given moment. One way to do that would be to set hook Deployments to use the Recreate update strategy, but that causes unnecessary downtime. Ironically, the ideal way to update a hook deployment would probably be to use a blue-green strategy with a Service that atomically switches to a fully-deployed set of replicas.

You can file an issue to track this if you like, but I don't think it's a high-priority problem for hooks. There may be some "flip-flopping" during rolling updates of hook code, but it should converge on its own once the rollout completes. Most hooks will probably work fine with just one replica, so they should update quickly.

manifests/metacontroller.yaml
crimsonfaith91: @enisoc Does Metacontroller still support v1.8? We will have GKE 1.11 GA rollout by mid-August (i.e. will not support v1.8 on GKE around a month later). Do we want to change the `apps/v1beta2` to `apps/v1` at that moment?

enisoc: I'd like to keep support for 1.8 until we have a compelling use case for relying on a feature that only exists in 1.9+.

crimsonfaith91:
I will squash if everything looks good.

enisoc:
@crimsonfaith91 Since we don't have automatic e2e tests yet, this is a good chance to make sure you can run the tests manually. Once you have a cluster set up, I recommend using the [Skaffold flow](https://metacontroller.app/contrib/build/#skaffold-build) to run your dev build. Since you just changed the Deployment to a StatefulSet, it's especially important to make sure you don't have any previous version of Metacontroller running; you can delete the `metacontroller` namespace first to ensure that.

I tried `skaffold run` locally and found one problem so far:

```
error: error validating "STDIN": error validating data: ValidationError(StatefulSet.spec): missing required field "serviceName" in io.k8s.api.apps.v1beta2.StatefulSetSpec; if you choose to ignore these errors, turn validation off with --validate=false
```

In this case, I think you can set `serviceName: ""` since we don't need per-pod DNS entries.

Once your dev build of Metacontroller is running, go into the `examples` dir and run `./test.sh` to manually kick off the smoke tests.

crimsonfaith91:
@enisoc I see. I thought it suffices to get the Travis CI build passed (the build does run unit tests). Thanks for the pointer!

crimsonfaith91:
@enisoc This PR is ready for review!

crimsonfaith91:
@enisoc The PR is ready for review!

manifests/dev/args.yaml
```diff
@@ -1,10 +1,11 @@
 # Override args for development mode.
-apiVersion: apps/v1beta1
-kind: Deployment
+apiVersion: apps/v1beta2
+kind: StatefulSet
 metadata:
   name: metacontroller
   namespace: metacontroller
 spec:
+  serviceName: ""
```
enisoc: Since `args.yaml` and `image.yaml` contain partial objects, used for kustomize overlays, we don't need to mention fields here if the value is the same as the base manifest.

PR #75, opened 2018-08-20
Expose workqueue stats

rlguarino:
This CL adds an http endpoint with one handler "/metrics" that serves
Prometheus metrics exported by MetaController. The metrics are collected
using OpenCensus and exported using the Prometheus HTTP exposition
format.

This CL also configures the Kubernetes workqueue package to collect its
statistics and expose them via the Prometheus http endpoint. We collect
and expose all of the workqueue metrics (Depth, Adds, Latency,
WorkDuration, Retries) tagged with the name of the queue.

### DIFF
--- a/main.go
+++ b/main.go
@@ -17,17 +17,23 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"flag"
+	"net/http"
 	"os"
 	"os/signal"
 	"sync"
 	"syscall"
 	"time"
 
+	"github.com/0xRLG/ocworkqueue"
 	"github.com/golang/glog"
+	"go.opencensus.io/exporter/prometheus"
+	"go.opencensus.io/stats/view"
 
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/util/workqueue"
 
 	"k8s.io/metacontroller/apis/metacontroller/v1alpha1"
 	mcclientset "k8s.io/metacontroller/client/generated/clientset/internalclientset"
@@ -42,6 +48,7 @@ import (
 var (
 	discoveryInterval = flag.Duration("discovery-interval", 30*time.Second, "How often to refresh discovery cache to pick up newly-installed resources")
 	informerRelist    = flag.Duration("cache-flush-interval", 30*time.Minute, "How often to flush local caches and relist objects from the API server")
+	debugAddr         = flag.String("debug-addr", ":9999", "The address to bind the debug http endpoints")
 )
 
 type controller interface {
@@ -54,6 +61,7 @@ func main() {
 
 	glog.Infof("Discovery cache flush interval: %v", *discoveryInterval)
 	glog.Infof("API server object cache flush interval: %v", *informerRelist)
+	glog.Infof("Debug http server address: %v", *debugAddr)
 
 	config, err := rest.InClusterConfig()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -94,6 +103,22 @@ func main() {
...
+	exporter, err := prometheus.NewExporter(prometheus.Options{})
+	if err != nil {
+		glog.Fatalf("can't create prometheus exporter: %v", err)
+	}
+	view.RegisterExporter(exporter)
+
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", exporter)
+	srv := &http.Server{
+		Addr:    ":9999",
```
rlguarino: @enisoc I want to get your opinion on how we should configure this. My initial guess would be to just use an environment variable, but I wanted to discuss.

rlguarino: I've added a flag to configure this.

main.go
```diff
@@ -17,17 +17,23 @@ limitations under the License.
...
+	"context"
 	"flag"
+	"net/http"
 	"os"
 	"os/signal"
 	"sync"
 	"syscall"
 	"time"
 
+	"github.com/0xRLG/ocworkqueue"
```
enisoc: Is there a benefit to going via OpenCensus instead of reusing the [Prometheus provider](https://sourcegraph.com/github.com/kubernetes/kubernetes/-/blob/pkg/util/workqueue/prometheus/prometheus.go) from core k8s? Whenever possible, I prefer to reuse code and patterns from the core since they're well-exercised.

main.go
```diff
@@ -42,6 +48,7 @@ import (
 var (
 	discoveryInterval = flag.Duration("discovery-interval", 30*time.Second, "How often to refresh discovery cache to pick up newly-installed resources")
 	informerRelist    = flag.Duration("cache-flush-interval", 30*time.Minute, "How often to flush local caches and relist objects from the API server")
+	debugAddr         = flag.String("debug-addr", "localhost:9999", "The address to bind the debug http endpoints")
```
enisoc: Binding to `localhost` by default means it won't be accessible from outside the Pod, doesn't it? Is that intentional? It looks like earlier you had just `":9999"`.

main.go
```diff
@@ -94,6 +105,22 @@ func main() {
...
+	view.RegisterExporter(exporter)
+
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", exporter)
+	srv := &http.Server{
+		Addr:    *debugAddr,
+		Handler: mux,
+	}
+	go func() {
+		glog.Fatalf("cannot serve debug endpoint: %v", srv.ListenAndServe())
```
enisoc: Maybe this should be logged but not fatal? The debug endpoint isn't strictly necessary for Metacontroller to perform its main duties. If I'm relying on Metacontroller to manage my app, I wouldn't want it to give up just because, for example, the metrics endpoint couldn't bind due to a port conflict.

In any case, we should probably ignore `ErrServerClosed` here because we call `Shutdown()` below.

rlguarino: You're right I remember fighting with this log line quite a bit I should have paid more attention. I'll change this to an `Errorf`

main.go
rlguarino: Yeah I think you're right I'll change it.

main.go
rlguarino: I didn't even know that existed I've been using this package for a long time now and I've only just made it open source. There isn't a huge difference between the two plain old prometheus and opencensus. Both result in Prometheus metrics being exposed on the debug endpoint. I think that OpenCensus ends up being cheaper to collect if you don't report it anywhere but we don't do that in this PR.

To be honest I think the fact that this project is in GoogleCloudPlatform and OpenCensus is being driven by Google people the two together made sense to me.

enisoc: Alright that makes sense. Let's just keep OpenCensus. This PR should be good to go then whenever you have a chance to make the other changes below.

rlguarino:
I addressed the comments and rebased this off master. I think the `GoPkg.lock` should now be the result of master's `GoPkg.locl` with my changes to `GoPkg.toml` and then running `dep ensure`. 

I ran:

```console
$ git checkout origin/master Gopkg.lock
$ rm -rf vendor/
$ dep ensure
```

PR #80, opened 2018-08-23
use new dynamic client

### LINKED ISSUE #47: Switch to new dynamic client
The old dynamic client is deprecated: https://github.com/kubernetes/kubernetes/pull/63446

This should probably wait until after 1.11 is released, so we know the new dynamic client is finalized.

### LINKED ISSUE #79: Unable to run `make deepcopy`
I'm unable to run `make` because the deepcopy target is failing.

```console
rssguar@cloudshell:~/gopath/src/k8s.io/metacontroller$ dep ensure --vendor-only
rssguar@cloudshell:~/gopath/src/k8s.io/metacontroller$ git s
rssguar@cloudshell:~/gopath/src/k8s.io/metacontroller$ make
+ Generating deepcopy funcs for metacontroller/v1alpha1
ERROR: logging before flag.Parse: F0822 17:04:10.405059     337 deepcopy.go:126] Failed loading boilerplate: open /home/rssguar/gopath/src/k8s.io/kubernetes/hack/boilerplate/boilerplate.go.txt: no such file or directory
goroutine 1 [running]:
k8s.io/metacontroller/vendor/github.com/golang/glog.stacks(0xc42000e000, 0xc42963dd40, 0xba, 0x10a)
        /home/rssguar/gopath/src/k8s.io/metacontroller/vendor/github.com/golang/glog/glog.go:769 +0xcf
k8s.io/metacontroller/vendor/github.com/golang/glog.(*loggingT).output(0x8d5fa0, 0xc400000003, 0xc429726a50, 0x8b5122, 0xb, 0x7e, 0x0)
        /home/rssguar/gopath/src/k8s.io/metacontroller/vendor/github.com/golang/glog/glog.go:720 +0x32d
k8s.io/metacontroller/vendor/github.com/golang/glog.(*loggingT).printf(0x8d5fa0, 0x3, 0x712bad, 0x1e, 0xc4270cfcf8, 0x1, 0x1)
        /home/rssguar/gopath/src/k8s.io/metacontroller/vendor/github.com/golang/glog/glog.go:655 +0x14b
k8s.io/metacontroller/vendor/github.com/golang/glog.Fatalf(0x712bad, 0x1e, 0xc4270cfcf8, 0x1, 0x1)
        /home/rssguar/gopath/src/k8s.io/metacontroller/vendor/github.com/golang/glog/glog.go:1148 +0x67
k8s.io/metacontroller/vendor/k8s.io/gengo/examples/deepcopy-gen/generators.Packages(0xc4296e8d80, 0xc4200a6960, 0x6e1f14, 0x6, 0xc4296e8d80)
        /home/rssguar/gopath/src/k8s.io/metacontroller/vendor/k8s.io/gengo/examples/deepcopy-gen/generators/deepcopy.go:126 +0xc5
k8s.io/metacontroller/vendor/k8s.io/gengo/args.(*GeneratorArgs).Execute(0xc4200a6960, 0xc420074db0, 0x6e1f14, 0x6, 0x71e060, 0x0, 0x0)
        /home/rssguar/gopath/src/k8s.io/metacontroller/vendor/k8s.io/gengo/args/args.go:193 +0x1b8
main.main()
        /home/rssguar/gopath/src/k8s.io/metacontroller/vendor/k8s.io/code-generator/cmd/deepcopy-gen/main.go:69 +0x1c4
Makefile:28: recipe for target 'deepcopy' failed
make: *** [deepcopy] Error 255
rssguar@cloudshell:~/gopath/src/k8s.io/metacontroller$ git show
commit 2a992e2bb526749a6f10cad9b570fbe0510599f5
Merge: 98b9014 8922193
Author: Anthony Yeh <enisoc@google.com>
Date:   Mon Jul 23 15:05:26 2018 -0700

    Merge pull request #66 from crimsonfaith91/one-mc

    only allow one Metacontroller instance to be active at a time

rssguar@cloudshell:~/gopath/src/k8s.io/metacontroller$ git status
```

Is this just me doing something strange? I was going to take a look at #47 but I can't confirm I'm in a working state to being with. I noticed we never actually run `make` in CI so it might not work anymore.

crimsonfaith91:
This PR addresses issue #47. It changes the codebase to use new dynamic client.

Fixes #47
Fixes #79


dynamic/informer/informer.go
```diff
@@ -105,10 +107,20 @@ type sharedResourceInformer struct {
 	close func()
 }
 
+type dynamicResourceClientListFunc func(metav1.ListOptions) (*unstructured.UnstructuredList, error)
+
+func castToCacheListFunc(fn dynamicResourceClientListFunc) cache.ListFunc {
+	options := metav1.ListOptions{}
```
crimsonfaith91: Is an empty `ListOptions` acceptable?

Gopkg.toml
```diff
@@ -30,19 +30,19 @@ required = [
...
   name = "k8s.io/code-generator"
-  version = "kubernetes-1.9.9"
+  version = "kubernetes-1.11.0"
 
 [[override]]
   name = "github.com/json-iterator/go"
-  version = "1.1.4" # as used by apimachinery@kubernetes-1.9.9
+  version = "1.1.5" # as used by apimachinery@kubernetes-1.11.0
 
 [prune]
```
rlguarino: I take it this is working for you? I only say that because I was running into all sorts of grouble getting code generation to work. See #79. I was able to get it to work adding the following to my `GoPkg.toml`:

```toml
[prune]
  go-tests = true
  unused-packages = true

  [[prune.project]]
    name = "k8s.io/code-generator"
    unused-packages = false
```

I was getting errors that the generator couldn't find the `boilerplate.go.txt` that comes with the code-generator repository.

dynamic/informer/informer.go
rlguarino: In practice, I've only set the LabelSelector of a ListOption before. It might be good hygiene to set timeout but that will likely cause more problems on slow clusters that it would prevent.

So, I think an empty LabelSelector is alright.

controller/composite/controller.go
```diff
@@ -471,7 +471,7 @@ func (pc *parentController) makeSelector(parent *unstructured.Unstructured, extr
 }
 
 func (pc *parentController) canAdoptFunc(parent *unstructured.Unstructured) func() error {
-	nsParentClient := pc.parentClient.WithNamespace(parent.GetNamespace())
+	nsParentClient := pc.parentClient
```
rlguarino: Do we actually need to reassign this variable here? Could we just directly call `ps.parentClient.Get(...)` below?

controller/composite/controller.go
```diff
@@ -536,7 +536,7 @@ func (pc *parentController) updateParentStatus(parent *unstructured.Unstructured
 	// We can't use Patch() because we need to ensure that the UID matches.
 	// TODO(enisoc): Use /status subresource when that exists.
 	// TODO(enisoc): Update status.observedGeneration when spec.generation starts working.
-	nsParentClient := pc.parentClient.WithNamespace(parent.GetNamespace())
+	nsParentClient := pc.parentClient
```
rlguarino: Simmarly to above about calling `pc.parentClient.UpdateWIthRetries()...`  directly instead of defining a new variable. We could use:

```
return pc.parentClient.UpdateWithRetries(...
```

crimsonfaith91: +1, you are right. :)

controller/composite/controller.go
crimsonfaith91: Agreed.

Gopkg.toml
crimsonfaith91: Me too - There is no `boilerplate.go.txt` within the vendor code-generator repository, so I added it manually (which is wrong) to unblock myself. I forgot to point this out, thanks for the reminder!

crimsonfaith91: Let me sync with @enisoc, and get back to you later. I believe your way of getting around this is better.

Gopkg.toml
```diff
@@ -30,19 +30,19 @@ required = [
...
 
 [[constraint]]
   name = "k8s.io/code-generator"
-  version = "kubernetes-1.9.9"
+  version = "kubernetes-1.11.0"
 
 [[override]]
   name = "github.com/json-iterator/go"
-  version = "1.1.4" # as used by apimachinery@kubernetes-1.9.9
+  version = "1.1.5" # as used by apimachinery@kubernetes-1.11.0
```
enisoc: How did you determine that apimachinery uses 1.1.5 at the kubernetes-1.11.0 tag? From the Godeps file (below) it looks like it still uses a commit that's contained in 1.1.4:

https://github.com/kubernetes/apimachinery/blob/103fd098999dc9c0c88536f5c9ad2e5da39373ae/Godeps/Godeps.json#L94-L95

That said, I'm fine updating to 1.1.5 if it works. We should just reword the comment to say `# same minor version track as used by apimachinery@kubernetes-1.11.0`

Gopkg.toml
enisoc: @crimsonfaith91 I agree with the solution @rlguarino suggested. Can you try adding that?

@rlguarino Thanks for tracking that down!

controller/composite/controller.go
enisoc: I agree too. In addition, make sure `pc.parentClient` is a namespace-agnostic client since a single `parentController` instance manages objects in multiple namespaces. In other words, the line below should ultimately read `pc.parentClient.Namespace(parent.GetNamespace()).Get(parent.GetName(), ...)`.

controller/composite/controller.go
enisoc: Also like above, this may need to change to account for `pc.parentClient` becoming namespace-agnostic. To make something like `pc.parentClient.Namespace(...).UpdateWithRetries(...)` work, we'll need to restructure how we add extra methods to ResourceInterface in our `dynamic/clientset` wrapper.

dynamic/clientset/dynamic_clientset.go
```diff
@@ -64,6 +64,15 @@ func (cs *Clientset) Kind(apiVersion, kind, namespace string) (*ResourceClient,
 	return cs.resource(apiResource, namespace)
 }
 
+func getResourceGVR(apiResource *dynamicdiscovery.APIResource) schema.GroupVersionResource {
```
enisoc: Let's move this to a method called `GroupVersionResource()` on `APIResource` in `dynamic/discovery`, to be consistent with `GroupVersionKind()` and make this immediately available anywhere we already have an object of this type.

dynamic/clientset/dynamic_clientset.go
```diff
@@ -64,6 +64,15 @@ func (cs *Clientset) Kind(apiVersion, kind, namespace string) (*ResourceClient,
...
+	return schema.GroupVersionResource{
+		Group:    resource.Group,
+		Version:  resource.Version,
+		Resource: resource.Name,
+	}
+}
+
 func (cs *Clientset) resource(apiResource *dynamicdiscovery.APIResource, namespace string) (*ResourceClient, error) {
 	// Create dynamic client for this apiVersion/resource.
 	gv := apiResource.GroupVersion()
```
enisoc: Can you check if the new dynamic client lets us leave GroupVersion unset? Ideally we should only need to call `dynamic.NewForConfig()` once in this package's `New()` function, and store that. I believe it was a limitation of the old dynamic client that we had to make a new one for different group-versions.

If you're able to remove the need to create a new client, you can also simplify this function by removing the `error` return.

dynamic/clientset/dynamic_clientset.go
```diff
@@ -64,6 +64,15 @@ func (cs *Clientset) Kind(apiVersion, kind, namespace string) (*ResourceClient,
...
+func getResourceGVR(apiResource *dynamicdiscovery.APIResource) schema.GroupVersionResource {
+	resource := apiResource.APIResource
+	return schema.GroupVersionResource{
+		Group:    resource.Group,
+		Version:  resource.Version,
+		Resource: resource.Name,
+	}
+}
+
 func (cs *Clientset) resource(apiResource *dynamicdiscovery.APIResource, namespace string) (*ResourceClient, error) {
```
enisoc: One of the improvements in the new dynamic client is that it decouples specification of the namespace from the creation of the client. We should mirror that by removing the `namespace` argument here and in `Clientset.Resource()`.

dynamic/clientset/dynamic_clientset.go
```diff
@@ -72,33 +81,27 @@ func (cs *Clientset) resource(apiResource *dynamicdiscovery.APIResource, namespa
...
+		resource:                       apiResource,
 	}, nil
 }
 
 type ResourceClient struct {
-	dynamic.ResourceInterface
+	dynamic.NamespaceableResourceInterface
 
-	dc       *dynamic.Client
+	dc       dynamic.Interface
```
enisoc: We might not need to store `dc` here anymore, since we no longer need to make a new client just to change the namespace.

dynamic/clientset/dynamic_clientset.go
```diff
@@ -72,33 +81,27 @@ func (cs *Clientset) resource(apiResource *dynamicdiscovery.APIResource, namespa
...
 	gv       schema.GroupVersion
 	resource *dynamicdiscovery.APIResource
 }
 
-func (rc *ResourceClient) WithNamespace(namespace string) *ResourceClient {
```
enisoc: In order to add `UpdateWithRetries()` for namespaced clients, we probably need to override `Namespace()` here and make it return a new wrapper struct that anonymously embeds a `dynamic.ResourceInterface` as well as implementing `UpdateWithRetries()`.

dynamic/informer/informer.go
enisoc: This is meant to just be an adapter function, so you should pass along the ListOptions from the caller. See my other comment below, which should make this clearer.

dynamic/informer/informer.go
```diff
@@ -105,10 +107,20 @@ type sharedResourceInformer struct {
...
+	return func(metav1.ListOptions) (runtime.Object, error) {
+		return runtime.Object(unstructuredList), err
+	}
+}
+
 func newSharedResourceInformer(client *dynamicclientset.ResourceClient, defaultResyncPeriod time.Duration, close func()) *sharedResourceInformer {
 	informer := cache.NewSharedIndexInformer(
 		&cache.ListWatch{
-			ListFunc:  client.List,
+			ListFunc:  castToCacheListFunc(client.List),
```
enisoc: Since we only need to do this once, I don't think it warrants all the stuff broken out above. It should be possible to write this concisely as something like this:

```go
ListFunc: func(opts metav1.ListOptions) (runtime.Object, error) {
  return client.List(opts)
}
```

Note that you don't need to cast the return value because a type that implements an interface is already assignable to a variable of that interface type.

Gopkg.toml
crimsonfaith91: @enisoc Sure! I will include the `[prune]` section.

controller/composite/controller.go
crimsonfaith91: Noted, thanks for pointing this out. :)

controller/composite/controller.go
crimsonfaith91: Thanks for the detailed explanation. Did you mean `ResourceClient`? `NamespaceableResourceInterface` is a vendor API.

Gopkg.toml
```diff
@@ -30,19 +30,19 @@ required = [
 
 [[constraint]]
   name = "k8s.io/client-go"
-  version = "~6.0"
+  revision = "kubernetes-1.11.0"
```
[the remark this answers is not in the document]
crimsonfaith91: Good point. I thought the requirement is to use the client-go version of 1.11 kubernetes version (misunderstood issue description).

dynamic/informer/informer.go
crimsonfaith91: This is a much better way to do it! Thanks for specifying that the casting is not necessary.

Gopkg.toml
crimsonfaith91: @enisoc How do you get the info that the commit is 1.1.4? The relevant PR should be [this](https://github.com/json-iterator/go/commit/f2b4162afba35581b6d4a50d3b8f34e33c144682), but tag 1.1.5 is in bold.

dynamic/clientset/dynamic_clientset.go
enisoc: Oh yeah, that one too. Good catch. :)

controller/composite/controller.go
enisoc: The original purpose of ResourceClient was to extend ResourceInterface with extra methods that are useful for Metacontroller, which works with many different resources from many different angles.

Now that the new dynamic client has been split into ResourceInterface and NamespaceableResourceInterface, we need to restructure ResourceClient to correspond to NamespaceableResourceInterface, and add another level of wrapper to extend ResourceInterface as well.

Gopkg.toml
crimsonfaith91: I see - we should take the earliest (not latest) tag. Thank you!
I verify that 1.1.5 works.

dynamic/clientset/dynamic_clientset.go
crimsonfaith91: Yes, the new dynamic client allows `GroupVersion` to be unset, but it does check correctness of the format. Refer [this](https://github.com/kubernetes/apimachinery/blob/master/pkg/runtime/schema/group_version.go#L215-L231) for more information.

I will remove `NewForConfig` and `error` return.

dynamic/clientset/dynamic_clientset.go
crimsonfaith91: Hasn't `UpdateWithRetries()` implemented already [here](https://github.com/GoogleCloudPlatform/metacontroller/blob/master/dynamic/clientset/dynamic_clientset.go#L125-L145)?

dynamic/clientset/dynamic_clientset.go
crimsonfaith91: Removal of `dc` means we have to find another way of initializing `ResourceClient.NamespaceableResourceInterface`. I will see what is the possible workaround. This will take me some time because I am basically have zero knowledge on API machinery.

dynamic/clientset/dynamic_clientset.go
```diff
@@ -33,75 +33,72 @@ import (
 type Clientset struct {
 	config    rest.Config
 	resources *dynamicdiscovery.ResourceMap
+	dc        dynamic.Interface
 }
 
-func New(config *rest.Config, resources *dynamicdiscovery.ResourceMap) *Clientset {
+func New(config *rest.Config, resources *dynamicdiscovery.ResourceMap) (*Clientset, error) {
+	// TODO: find a way to obtain APIResource for getting GroupVersion.
```
crimsonfaith91: @enisoc I shifted `dc` field from `ResourceClient` to `Clientset`. Not sure whether this is what you suggested. Feel free to sync by person if I get you wrong.

When calling `NewForConfig` within `New`, I couldn't find any way to obtain `apiVersion` and `resource` info, so I couldn't call `resources.Get(apiVersion, resource)` to obtain `the GroupVersion` for populating `config`.

dynamic/clientset/dynamic_clientset.go
enisoc: We need to implement UpdateWithRetries() twice, although they can both just call the same private function that does the real work. My goal was to match the pattern used in the new dynamic client for handling the fact that some API calls are namespaced and others are not.

Notice that [NamespaceableResourceInterface](https://godoc.org/k8s.io/client-go/dynamic#NamespaceableResourceInterface) contains both a `Namespace()` method *and* a full embedding of ResourceInterface:

```go
type NamespaceableResourceInterface interface {
    Namespace(string) ResourceInterface
    ResourceInterface
}
```

That means, for example, you can do both `dc.Resource().List()` and `dc.Resource().Namespace().List()`. Which one you use depends on whether you want to send a namespaced request or a cluster-scoped request. For example, you can list objects for one namespace, or for all namespaces.

To match this for the extra functions we add like UpdateWithRetries(), we should make sure both `rc.UpdateWithRetries()` and `rc.Namespace().UpdateWithRetries()` work, because we want to be able to update objects that are either cluster-scoped or namespaced.

dynamic/clientset/dynamic_clientset.go
crimsonfaith91: Thanks for verifying that I am on the right track. I will post log errors later if I could not resolve it. :)

dynamic/clientset/dynamic_clientset.go
crimsonfaith91: Oh, i see! I have to support rc.UpdateWithRetries() too for cluster-scoped operations.

dynamic/clientset/dynamic_clientset.go
```diff
@@ -147,3 +150,32 @@ func (rc *ResourceClient) UpdateWithRetries(orig *unstructured.Unstructured, upd
...
 }
+
+func (nrc *NamespacedResourceClient) GroupResource() schema.GroupResource {
+	return schema.GroupResource{
+		Group:    nrc.gv.Group,
+		Resource: nrc.resource.Name,
+	}
+}
+
+func (nrc *NamespacedResourceClient) UpdateWithRetries(orig *unstructured.Unstructured, update func(obj *unstructured.Unstructured) bool) (result *unstructured.Unstructured, err error) {
```
crimsonfaith91: I have to find a way to de-duplicate this function.

controller/common/manage_children.go
```diff
@@ -111,7 +111,7 @@ func deleteChildren(client *dynamicclientset.ResourceClient, parent *unstructure
 			// This observed object wasn't listed as desired.
 			glog.Infof("%v %v/%v: deleting %v %v", parent.GetKind(), parent.GetNamespace(), parent.GetName(), obj.GetKind(), obj.GetName())
 			uid := obj.GetUID()
-			err := client.Delete(name, &metav1.DeleteOptions{
+			err := client.Namespace(parent.GetNamespace()).Delete(name, &metav1.DeleteOptions{
```
crimsonfaith91: We can either set `Namespace` here, or earlier in the sync call tree via `dynClient.Kind(apiVersion, kind).Namespace(parent.GetNamespace())`. I prefer existing implementation because we may need to work with cluster-scoped `client` too.

dynamic/clientset/dynamic_clientset.go
```diff
@@ -33,70 +33,71 @@ import (
...
-	}
+func (cs *Clientset) resource(apiResource *dynamicdiscovery.APIResource) *ResourceClient {
 	return &ResourceClient{
-		ResourceInterface: dc.Resource(&apiResource.APIResource, namespace),
-		dc:                dc,
-		gv:                gv,
-		resource:          apiResource,
-	}, nil
+		NamespaceableResourceInterface: cs.dc.Resource(apiResource.GroupVersionResource()),
+		gv:                             apiResource.GroupVersion(),
```
enisoc: I don't remember why I originally stored `gv` in addition to `resource`, since you can get one from the other. Can you try removing `gv` from the struct and see if that simplifies things without adding some other cost?

dynamic/clientset/dynamic_clientset.go
```diff
@@ -143,3 +148,32 @@ func (rc *ResourceClient) UpdateWithRetries(orig *unstructured.Unstructured, upd
...
 }
+
+func (nrc *NamespacedResourceClient) GroupResource() schema.GroupResource {
+	return schema.GroupResource{
+		Group:    nrc.gv.Group,
+		Resource: nrc.resource.Name,
+	}
+}
+
+func (nrc *NamespacedResourceClient) UpdateWithRetries(orig *unstructured.Unstructured, update func(obj *unstructured.Unstructured) bool) (result *unstructured.Unstructured, err error) {
```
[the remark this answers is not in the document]
crimsonfaith91: Sound good to me, thanks!

dynamic/clientset/dynamic_clientset.go
crimsonfaith91: Noted, will try removing `gv`.

dynamic/clientset/dynamic_clientset.go
```diff
@@ -105,74 +99,57 @@ func (rc *ResourceClient) Kind() string {
 }
 
 func (rc *ResourceClient) GroupVersion() schema.GroupVersion {
-	return rc.gv
+	return rc.resource.GroupVersion()
 }
 
 func (rc *ResourceClient) GroupResource() schema.GroupResource {
```
crimsonfaith91: `GroupResource()` is still necessary for `ResourceClient` because it is called [here](https://github.com/GoogleCloudPlatform/metacontroller/blob/master/dynamic/informer/informer.go#L125).

controller/composite/controller.go
```diff
@@ -499,7 +498,7 @@ func (pc *parentController) claimChildren(parent *unstructured.Unstructured) (co
 	childMap := make(common.ChildMap)
 	for _, child := range pc.cc.Spec.ChildResources {
 		// List all objects of the child kind in the parent object's namespace.
-		childClient, err := pc.dynClient.Resource(child.APIVersion, child.Resource, namespace)
+		childClient, err := pc.dynClient.Resource(child.APIVersion, child.Resource)
```
enisoc: Sorry for not noticing this earlier: I think since we no longer set the namespace here, we need to fix `UnstructuredManager` in the `dynamic/controllerref` package to call `Namespace()` when needed (if the child object has a namespace set).

crimsonfaith91: @enisoc Yes, they passed. I ran all the tests for every commit now with image containing my changes via `skaffold run` after changing `skaffold.yaml` and `image.yaml` to my docker ID.

crimsonfaith91: Noted, I will change `dynamic/controllerref` package accordingly.

crimsonfaith91: @enisoc Do we also need to fix `ControllerRevisionManager`? `ControllerRevision` are namespaced objects.

enisoc: That's a good point, I hadn't thought of that. I just checked though and it looks like ControllerRevisionManager is unaffected by your changes because we use a generated client for that rather than the dynamic client.

crimsonfaith91: I see. I will push a commit soon. Thanks!

crimsonfaith91:
@enisoc Should we combine these [two](https://github.com/GoogleCloudPlatform/metacontroller/blob/master/dynamic/controllerref/unstructured.go#L71-L77) if-blocks into one if-else-block?

enisoc:
Style-wise, I prefer them as separate blocks. The first block follows the pattern of short-circuiting control flow (return/break/continue) if an error occurs. The second block is not part of error handling, but is actually part of the [normal flow](https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow) -- it's just that the normal flow happens to consist of an if block.

PR #87, opened 2018-09-05
mount certificates path

### LINKED ISSUE #86: x509: failed to load system roots and no roots provided
I've created a simple decorator with an external (not a k8s service) webhook. I'm getting the x509 error. 

Potential solution: mount /etc/ssl/certs in metacontroller container.

```
...
        volumeMounts:
        - mountPath: /etc/ssl/certs
          name: ca-certificates
          readOnly: true
    volumes:
    - name: ca-certificates
       hostPath:
          path: /etc/ssl/certs
```

lionelvillard:
Fixes #86

### DIFF
--- a/manifests/metacontroller.yaml
+++ b/manifests/metacontroller.yaml
@@ -70,3 +70,11 @@ spec:
         - --logtostderr
         - -v=4
         - --discovery-interval=20s
+        volumeMounts:
+        - mountPath: /etc/ssl/certs
+          name: ca-certs
+          readOnly: true
+      volumes:
+      - hostPath:
+          path: /etc/ssl/certs
+        name: ca-certs


manifests/metacontroller.yaml
```diff
@@ -70,3 +70,11 @@ spec:
         - --logtostderr
         - -v=4
         - --discovery-interval=20s
+        volumeMounts:
+        - mountPath: /etc/ssl/certs
+          name: ca-certs
+          readOnly: true
+      volumes:
+      - hostPath:
+          path: /etc/ssl/certs
```
rlguarino: I'm not an expert on how every distro lays out it's certificates but I'm sure they're not always the same.

manifests/metacontroller.yaml
```diff
@@ -70,3 +70,11 @@ spec:
...
+        name: ca-certs
```
rlguarino: There should probably be a `type` here. The default empty string exists for backwards compatibility, but we should be more explicit about what checks we need.

rlguarino:
@lionelvillard can you take a look & test #88 and let me know if that will fix the issue?

enisoc:
It sounds like this can be closed since #88 fixes the original issue #86.

Someone proposed #50 for specifying one default, custom CA for the whole Metacontroller server. If anyone needs that, we could pick up where that PR left off. At some point, we may also need per-hook CA config as @rlguarino suggested.

PR #95, opened 2018-09-21
Simpify ResourceClient namespace handling.

enisoc:
The main goal is to remove the duplication caused by introducing `NamespacedResourceClient`, because I want to add more methods similar to UpdateWithRetries as part of the work on finalizers, and I don't want to have to make extra wrappers for every new method.

We introduced `NamespacedResourceClient` to preserve the expected semantic from the generated clients that you can't chain calls to `.Namespace()`. However, given the dynamic nature of ResourceClient, I think it's simpler if we just keep the one type and allow chaining.

As part of that, I added a check that makes calling `.Namespace()` revert to a cluster-scoped client if the given namespace is empty, or if we know from discovery info that the resource is cluster-scoped. This makes it safer to just always call `Namespace()` throughout, and it will just do the right thing for either cluster-scoped or namespace-scoped resources.

I also renamed UpdateWithRetries to AtomicUpdate and added comments to clarify what it's for, and cleaned up some places where we now need to deal with objects that may or may not have namespaces.

### DIFF
--- a/dynamic/clientset/clientset.go
+++ b/dynamic/clientset/clientset.go
@@ -22,7 +22,6 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/util/retry"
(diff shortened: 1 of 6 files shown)


dynamic/clientset/clientset.go
```diff
@@ -71,85 +70,82 @@ func (cs *Clientset) Kind(apiVersion, kind string) (*ResourceClient, error) {
...
-	return rc.resource.GroupVersion()
-}
-
-func (rc *ResourceClient) GroupResource() schema.GroupResource {
-	return schema.GroupResource{
-		Group:    rc.resource.GroupVersion().Group,
-		Resource: rc.resource.Name,
+// Namespace returns a copy of the ResourceClient with the client namespace set.
+//
+// This can be chained to set the namespace to something else.
```
crimsonfaith91: Does it mean that `.Namespace("a").Namespace("b")` will have the namespace set as `b`?

dynamic/clientset/clientset.go
```diff
@@ -71,85 +70,82 @@ func (cs *Clientset) Kind(apiVersion, kind string) (*ResourceClient, error) {
...
-
-func updateWithRetries(ri dynamic.ResourceInterface, resource *dynamicdiscovery.APIResource, orig *unstructured.Unstructured, update func(obj *unstructured.Unstructured) bool) (result *unstructured.Unstructured, err error) {
+// AtomicUpdate performs an atomic read-modify-write loop, retrying on
+// optimistic concurrency conflicts.
+//
+// It only uses the identity (name/namespace/uid) of the provided 'orig' object,
+// not the contents. The object passed to the update() func will be from a live
+// GET against the API server.
+//
+// This should only be used for unconditional writes, as in, "I want to make
```
crimsonfaith91: Did previous `UpdateWithRetries` usages mean to be unconditional writes?

crimsonfaith91:
I reviewed each line carefully, and thought all changes were made in the best possible way. :)
Thanks for pointing me to this PR, really learnt a lot from it!

crimsonfaith91:
@enisoc I have LGTM-ed this PR last week. This should be good to merge.

dynamic/clientset/clientset.go
enisoc: Yes

dynamic/clientset/clientset.go
enisoc: Yes, we should only have been using UpdateWithRetries in cases when we *don't* want the write to be conditional on the ResourceVersion being the same as our last read.

crimsonfaith91: Noted, thanks!

PR #96, opened 2018-09-21
add scope on crd spec from guide / Create a Controller

code0x9:
CRD helloworlds.example.com spec needs 'scope' on spec. without scope:

`error: error validating "crd.yaml": error validating data: ValidationError(CustomResourceDefinition.spec): missing required field "scope" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1beta1.CustomResourceDefinitionSpec; if you choose to ignore these errors, turn validation off with --validate=false
`

### DIFF
--- a/docs/_guide/create.md
+++ b/docs/_guide/create.md
@@ -51,6 +51,7 @@ spec:
     kind: HelloWorld
     plural: helloworlds
     singular: helloworld
+  scope: Namespaced
 ```
 
 Then apply it to your cluster:


enisoc:
Out of curiosity, can you share what kubectl/cluster version this happened with? With the versions I use, the `scope` field has become optional and defaults to Namespaced.

In any case, it's best to add it if it's needed to support older clients. Thanks for the fix!

code0x9:
here's my k8s versions
```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.1", GitCommit:"b1b29978270dc22fecc592ac55d903350454310a", GitTreeState:"clean", BuildDate:"2018-07-18T11:37:06Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.2", GitCommit:"bb9ffb1654d4a729bb4cec18ff088eacc153c239", GitTreeState:"clean", BuildDate:"2018-08-07T23:08:19Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"linux/amd64"}
```
I'm using rancher 2.0.8. each k8s node is using ubuntu 16.04 LTS provisioned by https://github.com/rancher/install-docker/blob/master/17.03.2.sh

code0x9:
I think this is the cause of validation error: https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apis/apiextensions/validation/validation.go#L115

PR #100, opened 2018-09-25
Add finalize hook

### LINKED ISSUE #60: Execute custom cleanup code on deletion with finalizers
The use case is what I have discussed with you @enisoc in #58. We (@folago and me) are trying to implement a LB controller based on `Service` resource using `DecoratorController`. We create an external LB when a new service is created and pass the info using annotations on the `Service` resource. But when the `Service` resource gets deleted, we need to have this info passed to our controller, so we can delete the LB entry for this service as well. As the [code](https://github.com/GoogleCloudPlatform/metacontroller/blob/master/controller/decorator/controller.go#L249) stands now, it is only possible to delete the objects in the cluster using ownerreference, but objects created outside the cluster is not possible to clean up. It would be good to have this as a configurable option in the controller spec and default disabled, if preferred, to preserve the current behavior.

enisoc:
This adds a `finalize` hook to both CompositeController and DecoratorController.

If a `finalize` hook is defined, Metacontroller adds a controller-specific finalizer to `metadata.finalizers` to block deletion of the parent object until the user's `finalize` handler reports that it's satisfied. This enables reliable cleanup of state that the built-in garbage collector can't handle (e.g. resources in an external system).

Closes #60


enisoc:
@rlguarino @crimsonfaith91 This is ready for review now. You might want to start with the docs and examples to get an overview of the intended UX before diving into the implementation.

docs/_api/compositecontroller.md
```diff
@@ -392,3 +394,90 @@ That is, you should [set only the fields that you care about](/api/apply/).
 Note that your webhook handler must return a response with a status code of `200`
 to be considered successful. Metacontroller will wait for a response for up to the
 amount defined in the [Webhook spec](/api/hook/#webhook).
+
+### Finalize Hook
```
crimsonfaith91: Doc section LGTM-ed. I may post several questions here (for inquiry / verification purpose) here later when reviewing the codes.

examples/service-per-pod/README.md
```diff
@@ -39,6 +39,9 @@ spec:
 ...
 ```
 
+If the StatefulSet is then deleted, or if the `service-per-pod-label` annotation
+is removed to opt out of the decorator, any Services created will be claned up.
```
crimsonfaith91: nit: cleaned

examples/catset/test.sh
```diff
@@ -38,3 +40,16 @@ kubectl patch $cs nginx-backend --type=merge -p '{"spec":{"replicas":2}}'
...
 echo "Wait for 2 Pods to be Ready..."
 until [[ "$(kubectl get $cs nginx-backend -o 'jsonpath={.status.readyReplicas}')" -eq 2 ]]; do sleep 1; done
+
+echo "Append our own finalizer so we can read the final state..."
+kubectl patch $cs nginx-backend --type=json -p '[{"op":"add","path":"/metadata/finalizers/-","value":"'${finalizer}'"}]'
+
+echo "Delete CatSet..."
+kubectl delete $cs nginx-backend --wait=false
+
+echo "Expect CatSet's finalizer to scale to 0..."
```
crimsonfaith91: Expect CatSet's number of replicas to scale to 0?

dynamic/object/metadata.go
```diff
@@ -0,0 +1,52 @@
...
+	for _, item := range obj.GetFinalizers() {
+		if item == name {
+			return true
+		}
+	}
+	return false
+}
+
+// AddFinalizer adds the named finalizer to obj, if it isn't already present.
+func AddFinalizer(obj metav1.Object, name string) {
```
crimsonfaith91: This function can return a `bool` value, in which `true` indicates the finalizer has been set, and `false` if the finalizer already exists (no-op).

(But I think you have reason for not doing so - most probably due to coding style.)

dynamic/clientset/clientset.go
```diff
@@ -150,3 +151,27 @@ func (rc *ResourceClient) AtomicUpdate(orig *unstructured.Unstructured, update f
 	})
 	return result, err
 }
+
+// AddFinalizer adds the given finalizer to the list, if it isn't there already.
+func (rc *ResourceClient) AddFinalizer(orig *unstructured.Unstructured, name string) (*unstructured.Unstructured, error) {
+	return rc.AtomicUpdate(orig, func(obj *unstructured.Unstructured) bool {
+		if dynamicobject.HasFinalizer(obj, name) {
```
crimsonfaith91: This `if` clause can be skipped if it turns out that it is fine for `AddFinalizer` to return a `bool` value (refer below).

crimsonfaith91:
@enisoc I reviewed all commits except the main commit (5th commit). Should we also add finalizer smoke tests for `BlueGreen` and `IndexedJob`?

dynamic/clientset/clientset.go
enisoc: I thought about that, but ultimately I felt like the call sites for `AddFinalizer` will be more readable if they explicitly call `HasFinalizer` when it matters, as is done here. It's an extra loop pass over the list, but I have no data pointing to this as a performance concern so I chose readability over brevity or premature optimization.

dynamic/object/metadata.go
enisoc: We can discuss this in the thread above.

examples/catset/test.sh
enisoc: The implicit meaning was intended to be, "Expect CatSet's finalizer to scale the CatSet to 0 replicas". I'll update the comment to add back the elided words for clarity.

examples/service-per-pod/README.md
enisoc: Done. Thanks!

examples/service-per-pod/test.sh
```diff
@@ -24,3 +27,30 @@ until [[ "$(kubectl get svc nginx-2 -o 'jsonpath={.spec.selector.pod-name}')" ==
 
 echo "Wait for pod-name label..."
 until [[ "$(kubectl get pod nginx-2 -o 'jsonpath={.metadata.labels.pod-name}')" == "nginx-2" ]]; do sleep 1; done
+
+echo "Remove annotation to opt out of service-per-pod without deleting the StatefulSet..."
+kubectl annotate statefulset nginx service-per-pod-label-
```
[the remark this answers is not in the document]
enisoc: For `kubectl annotate`, the syntax `annotation-` means "remove this annotation", while `annotation=` means "set this annotation to the empty string". Since our decorator selector only requires this label to exist, we need to remove it entirely to prevent a match.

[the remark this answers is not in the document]
crimsonfaith91: noted, never know this. thanks!

dynamic/clientset/clientset.go
crimsonfaith91: noted!

controller/common/finalizer/finalizer.go
```diff
@@ -0,0 +1,79 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	dynamicclientset "k8s.io/metacontroller/dynamic/clientset"
+	dynamicobject "k8s.io/metacontroller/dynamic/object"
+)
+
+// Manager encapsulates controller logic for dealing with finalizers.
+type Manager struct {
```
crimsonfaith91: nit: `FinalizerManager`

controller/composite/hooks.go
```diff
@@ -31,20 +31,44 @@ type syncHookRequest struct {
...
 	}
+
 	var response syncHookResponse
-	if err := hooks.Call(cc.Spec.Hooks.Sync, request, &response); err != nil {
-		return nil, fmt.Errorf("sync hook failed: %v", err)
+
+	// First check if we should instead call the finalize hook,
+	// which has the same API as the sync hook except that it's
+	// called while the object is pending deletion.
+	if request.Parent.GetDeletionTimestamp() != nil && cc.Spec.Hooks.Finalize != nil {
```
crimsonfaith91: Is it possible to factor out common functions for both composite and decorator controllers?

crimsonfaith91:
/lgtm
Great work!

controller/composite/hooks.go
enisoc: It would be nice, but I'm not sure where we could draw a clean line of abstraction. For example, the logic on this particular line is a little bit different than in DecoratorController's version of `callSyncHook`.

PR #102, opened 2018-09-27
Support status subresource and ObservedGeneration

### LINKED ISSUE #26: Support status subresource and ObservedGeneration.
Now that we have support for /status in both [CRD](https://github.com/kubernetes/kubernetes/pull/55168) and the [dynamic client](https://github.com/kubernetes/kubernetes/pull/56717), we should use /status whenever it's available.

As of k8s 1.10, /status for CRD is behind an alpha feature gate, and is also optional on a per-CRD basis. If we neglect to use /status when it's available, our status update will be silently ignored. If we try to use /status when it's not available, our status update will fail. We should be able to tell whether /status is available for a given resource by looking at discovery, without requiring the user to configure anything on the Metacontroller side (they'll just enable /status in the CRD itself).

We should also set `status.observedGeneration` now that CRD is capable of incrementing `metadata.generation` (if the /status subresource is enabled for that CRD).

https://github.com/kstmp/metacontroller/blob/8c2b6730d8db89c5eba31bf180ebd9fbe93b3a5c/controller/composite/controller.go#L256-L271

Lastly, for controllers that support ObservedGeneration, we can consider an optimization to avoid syncing if the controller object is updated, but the spec hasn't changed (ObservedGeneration = Generation). For example, we currently might trigger a sync as a result of updating our own status.

crimsonfaith91:
Closes #26

Items achieved:
1. added an `AtomicUpdateStatus` that calls `Update()` or `UpdateStatus` based on existence of status subresource
2. added a `.subresources` field for `APIResource` to record all subresources of a given resource (not just status subresource)
3. added a `HasSubresource` function for `APIResource` to determine whether a subresource exists
4. revert `.status` field for `ApplyUpdate()`

All smoke tests pass in latest commit.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -16,8 +16,8 @@
 [[projects]]
   name = "github.com/davecgh/go-spew"
   packages = ["spew"]
-  revision = "346938d642f2ec3594ed81d874461961cd0faa76"
-  version = "v1.1.0"
+  revision = "8991bc29aa16c548c550c7ff78260e27b9ab7c73"
+  version = "v1.1.1"
 
 [[projects]]
   name = "github.com/ghodss/yaml"
@@ -31,8 +31,8 @@
     "proto",
     "sortkeys"
   ]
-  revision = "1adfc126b41513cc696b209667c8656ea7aac67c"
-  version = "v1.0.0"
+  revision = "636bf0302bc95575d69441b25a2603156ffdddf1"
+  version = "v1.1.1"
 
 [[projects]]
   branch = "master"
@@ -49,24 +49,24 @@
     "ptypes/duration",
     "ptypes/timestamp"
   ]
-  revision = "925541529c1fa6821df4e44ce2723319eb2be768"
-  version = "v1.0.0"
+  revision = "aa810b61a9c79d51363740d207bb46cf8e620ed5"
+  version = "v1.2.0"
 
 [[projects]]
   branch = "master"
   name = "github.com/google/btree"
   packages = ["."]
-  revision = "e89373fe6b4a7413d7acd6da1725b83ef713e6e4"
+  revision = "4030bb1f1f0c35b30ca7009e9ebd06849dd45306"
 
 [[projects]]
-  branch = "master"
   name = "github.com/google/go-jsonnet"
   packages = [
     ".",
     "ast",
     "parser"
   ]
-  revision = "b0459e4867cd151f426eebf22bafe08c7eb608bf"
+  revision = "6144c57d2a054b72ff46219f655bf863b940e174"
+  version = "v0.11.2"
 
 [[projects]]
   branch = "master"
@@ -81,8 +81,8 @@
     "compiler",
     "extensions"
   ]
-  revision = "ee43cbb60db7bd22502942cccbc39059117352ab"
-  version = "v0.1.0"
+  revision = "7c663266750e7d82587642f65e60bc4083f1f84e"
+  version = "v0.2.0"
 
 [[projects]]
   branch = "master"
@@ -94,19 +94,19 @@
   revision = "9cad4c3443a7200dd6400aef47183728de563a38"
 
 [[projects]]
-  branch = "master"
   name = "github.com/hashicorp/golang-lru"
   packages = [
     ".",
     "simplelru"
   ]
-  revision = "0fb14efe8c47ae851c0034ed7a448854d3d34cf3"
+  revision = "20f1fb78b0740ba8c3cb143a61e86ba5c8669768"
+  version = "v0.5.0"
 
 [[projects]]
   name = "github.com/json-iterator/go"
   packages = ["."]
   revision = "1624edc4454b8682399def8740d46db5e4362ba4"
-  version = "1.1.5"
+  version = "v1.1.5"
 
 [[projects]]
   name = "github.com/matttproud/golang_protobuf_extensions"
@@ -171,13 +171,13 @@
     "nfs",
     "xfs"
   ]
-  revision = "05ee40e3a273f7245e8777337fc7b46e533a9a92"
+  revision = "185b4288413d2a0dd0806f78c90dde719829e5ae"
 
 [[projects]]
   name = "github.com/spf13/pflag"
   packages = ["."]
-  revision = "e57e3eeb33f795204c1ca35f56c44f83227c6e66"
-  version = "v1.0.0"
+  revision = "298182f68c66c05229eb03ac171abe6e309ee79a"
+  version = "v1.0.3"
 
 [[projects]]
   name = "go.opencensus.io"
@@ -198,19 +198,19 @@
   branch = "master"
   name = "golang.org/x/crypto"
   packages = ["ssh/terminal"]
-  revision = "614d502a4dac94afa3a6ce146bd1736da82514c6"
+  revision = "7c1a557ab941a71c619514f229f0b27ccb0c27cf"
 
 [[projects]]
   branch = "master"
   name = "golang.org/x/net"
   packages = [
     "context",
+    "http/httpguts",
     "http2",
     "http2/hpack",
-    "idna",
-    "lex/httplex"
+    "idna"
   ]
-  revision = "e0c57d8f86c17f0724497efcb3bc617e82834821"
+  revision = "146acd28ed5894421fb5aac80ca93bc1b1f46f87"
 
 [[projects]]
   branch = "master"
@@ -219,7 +219,7 @@
     "unix",
     "windows"
   ]
-  revision = "1a700e749ce29638d0bbcb531cce1094ea096bd3"
+  revision = "4497e2df6f9e69048a54498c7affbbec3294ad47"
 
 [[projects]]
   name = "golang.org/x/text"
@@ -256,19 +256,19 @@
     "imports",
     "internal/fastwalk"
   ]
-  revision = "d600f31f815368e57b551e524b0169bd175f7edb"
+  revision = "29f11e2b93f4d66f7c335bd7b2892836d4944f5c"
 
 [[projects]]
   name = "gopkg.in/inf.v0"
   packages = ["."]
-  revision = "3887ee99ecf07df5b447e9b00d9c0b2adaa9f3e4"
-  version = "v0.9.0"
+  revision = "d2d2541c53f18d2a059457998ce2876cc8e67cbf"
+  version = "v0.9.1"
 
 [[projects]]
   name = "gopkg.in/yaml.v2"
   packages = ["."]
-  revision = "7f97868eec74b32b0982dd158a51a446d1da7eb5"
-  version = "v2.1.1"
+  revision = "5420a8b6744d3b0345ab293f6fcba19c978f1183"
+  version = "v2.2.1"
 
 [[projects]]
   branch = "master"
@@ -304,7 +304,7 @@
     "storage/v1alpha1",
     "storage/v1beta1"
   ]
-  revision = "2315b41a07e851014d77070954fc391a9b2d9aa9"
+  revision = "e3c5c37695fbce2b0b46845e3bb4806c4b8faf47"
 
 [[projects]]
   name = "k8s.io/apimachinery"
@@ -421,13 +421,13 @@
     "parser",
     "types"
   ]
-  revision = "fdcf9f9480fdd5bf2b3c3df9bf4ecd22b25b87e2"
+  revision = "4242d8e6c5dba56827bb7bcf14ad11cda38f3991"
 
 [[projects]]
   name = "k8s.io/kubernetes"
   packages = ["pkg/util/pointer"]
-  revision = "bee2d1505c4fe820744d26d41ecd3fdd4a3d6546"
-  version = "v1.9.4"
+  revision = "a4529464e4629c21224b3d52edfe0ea91b072862"
+  version = "v1.11.3"
 
 [solve-meta]
   analyzer-name = "dep"
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,233 +2,184 @@
 
 
 [[projects]]
-  digest = "1:3dcf9df58e425f7ace052d832e683930533be49fcde3b6d66cebb2ed5a97bb3f"
   name = "github.com/0xRLG/ocworkqueue"
   packages = ["."]
-  pruneopts = "UT"
   revision = "40aa506a5a3c9f494b65cf929bb34b6ab1d02617"
   version = "0.1.0"
 
 [[projects]]
   branch = "master"
-  digest = "1:d6afaeed1502aa28e80a4ed0981d570ad91b2579193404256ce672ed0a609e0d"
   name = "github.com/beorn7/perks"
   packages = ["quantile"]
-  pruneopts = "UT"
   revision = "3a771d992973f24aa725d07868b467d1ddfceafb"
 
 [[projects]]
-  digest = "1:a2c1d0e43bd3baaa071d1b9ed72c27d78169b2b269f71c105ac4ba34b1be4a39"
   name = "github.com/davecgh/go-spew"
   packages = ["spew"]
-  pruneopts = "UT"
   revision = "346938d642f2ec3594ed81d874461961cd0faa76"
   version = "v1.1.0"
 
 [[projects]]
-  digest = "1:2cd7915ab26ede7d95b8749e6b1f933f1c6d5398030684e6505940a10f31cfda"
   name = "github.com/ghodss/yaml"
   packages = ["."]
-  pruneopts = "UT"
   revision = "0ca9ea5df5451ffdf184b4428c902747c2c11cd7"
   version = "v1.0.0"
 
 [[projects]]
-  digest = "1:a7534feda0f15b5fd691e59e4fb6b7547e27df4b415a62e02c7cb71b3439c1b1"
   name = "github.com/gogo/protobuf"
   packages = [
     "proto",
-    "sortkeys",
+    "sortkeys"
   ]
-  pruneopts = "UT"
   revision = "1adfc126b41513cc696b209667c8656ea7aac67c"
   version = "v1.0.0"
 
 [[projects]]
   branch = "master"
-  digest = "1:1ba1d79f2810270045c328ae5d674321db34e3aae468eb4233883b473c5c0467"
   name = "github.com/golang/glog"
   packages = ["."]
-  pruneopts = "UT"
   revision = "23def4e6c14b4da8ac2ed8007337bc5eb5007998"
 
 [[projects]]
-  digest = "1:8caffcd8995b0eae7d2c18b2baefabef331bb05b1e16ed2b26b732c3349e6989"
   name = "github.com/golang/protobuf"
   packages = [
     "proto",
     "ptypes",
     "ptypes/any",
     "ptypes/duration",
-    "ptypes/timestamp",
+    "ptypes/timestamp"
   ]
-  pruneopts = "UT"
   revision = "925541529c1fa6821df4e44ce2723319eb2be768"
   version = "v1.0.0"
 
 [[projects]]
   branch = "master"
-  digest = "1:9887333bbef17574b1db5f9893ea137ac44107235d624408a3ac9e0b98fbb2cb"
   name = "github.com/google/btree"
   packages = ["."]
-  pruneopts = "UT"
   revision = "e89373fe6b4a7413d7acd6da1725b83ef713e6e4"
 
 [[projects]]
   branch = "master"
-  digest = "1:9e21a98615890056a2b052e94035c87b21f8fea93d01aeeebae605fcc398b3e4"
   name = "github.com/google/go-jsonnet"
   packages = [
     ".",
     "ast",
-    "parser",
+    "parser"
   ]
-  pruneopts = "UT"
   revision = "b0459e4867cd151f426eebf22bafe08c7eb608bf"
 
 [[projects]]
   branch = "master"
-  digest = "1:3ee90c0d94da31b442dde97c99635aaafec68d0b8a3c12ee2075c6bdabeec6bb"
   name = "github.com/google/gofuzz"
   packages = ["."]
-  pruneopts = "UT"
   revision = "24818f796faf91cd76ec7bddd72458fbced7a6c1"
 
 [[projects]]
-  digest = "1:5e031a35b76ee001fa9ca9d598298054a123d080e00d13a8dafcfc5e3ecd5b58"
   name = "github.com/googleapis/gnostic"
   packages = [
     "OpenAPIv2",
     "compiler",
-    "extensions",
+    "extensions"
   ]
-  pruneopts = "UT"
   revision = "ee43cbb60db7bd22502942cccbc39059117352ab"
   version = "v0.1.0"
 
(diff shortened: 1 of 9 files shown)


controller/composite/controller.go
```diff
@@ -42,6 +42,7 @@ import (
 	dynamicdiscovery "k8s.io/metacontroller/dynamic/discovery"
 	dynamicinformer "k8s.io/metacontroller/dynamic/informer"
 	k8s "k8s.io/metacontroller/third_party/kubernetes"
+	// dynamicobject "k8s.io/metacontroller/dynamic/object"
```
crimsonfaith91: I will remove this later.

dynamic/clientset/clientset.go
```diff
@@ -150,3 +150,28 @@ func (rc *ResourceClient) AtomicUpdate(orig *unstructured.Unstructured, update f
...
+		}
+		if changed := update(current); !changed {
+			// There's nothing to do.
+			result = current
+			return nil
+		}
+		result, err = rc.UpdateStatus(current)
+		return err
+	})
+	fmt.Printf("\n\n4\n\n")
```
crimsonfaith91: This too.

crimsonfaith91:
@enisoc I wish to ensure I am on the right track. Thanks!

controller/composite/controller.go
```diff
@@ -232,17 +233,25 @@ func (pc *parentController) enqueueParentObject(obj interface{}) {
...
-	// Ignore updates where the ResourceVersion changed (not a resync)
-	// but the spec hasn't changed (e.g. our own status updates).
-	// TODO(metacontroller#26): Use ObservedGeneration instead of this.
+	// Ignore updates if the spec hasn't changed (e.g. our own status updates).
 	oldParent := old.(*unstructured.Unstructured)
 	curParent := cur.(*unstructured.Unstructured)
-	if curParent.GetResourceVersion() != oldParent.GetResourceVersion() {
+	oldStatus := k8s.GetNestedField(oldParent.UnstructuredContent(), "status")
+
+	// Status subresource is not found, so it means the subresource is not supported.
```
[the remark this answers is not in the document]
crimsonfaith91: Yes, this part has to be replaced with `ResourceMap` check. I didn't add a TODO here because this PR is already marked as a WIP. We should add TODOs if we want this PR to merge without incorporating changes covered by the TODOs.

controller/composite/controller.go
```diff
@@ -530,18 +539,31 @@ func (pc *parentController) claimChildren(parent *unstructured.Unstructured) (co
...
-	})
+	}
+}
+
+func (pc *parentController) updateParentStatus(parent *unstructured.Unstructured, status map[string]interface{}) (*unstructured.Unstructured, error) {
+	// Overwrite .status field of parent object without touching other parts.
+	// We can't use Patch() because we need to ensure that the UID matches.
+
+	// If status subresource exists, call AtomicStatusUpdate().
+	if pc.resources.Get(pc.parentResource.APIVersion, "status") != nil {
```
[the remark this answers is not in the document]
crimsonfaith91: done

controller/composite/controller.go
```diff
@@ -530,18 +539,31 @@ func (pc *parentController) claimChildren(parent *unstructured.Unstructured) (co
...
+		return pc.parentClient.Namespace(parent.GetNamespace()).
+			AtomicStatusUpdate(parent, updateParentStatusFunc(status))
```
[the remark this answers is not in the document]
crimsonfaith91: done

controller/composite/controller.go
```diff
@@ -530,18 +539,31 @@ func (pc *parentController) claimChildren(parent *unstructured.Unstructured) (co
...
+func updateParentStatusFunc(status map[string]interface{}) func(*unstructured.Unstructured) bool {
+	return func(obj *unstructured.Unstructured) bool {
 		oldStatus := k8s.GetNestedField(obj.UnstructuredContent(), "status")
 		if reflect.DeepEqual(oldStatus, status) {
 			// Nothing to do.
 			return false
 		}
+
 		k8s.SetNestedField(obj.UnstructuredContent(), status, "status")
+		k8s.SetNestedField(obj.UnstructuredContent(), obj.GetGeneration(), "status", "observedGeneration")
```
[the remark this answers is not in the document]
crimsonfaith91: done

controller/composite/controller.go
```diff
@@ -232,17 +233,25 @@ func (pc *parentController) enqueueParentObject(obj interface{}) {
 }
 
 func (pc *parentController) updateParentObject(old, cur interface{}) {
-	// Ignore updates where the ResourceVersion changed (not a resync)
-	// but the spec hasn't changed (e.g. our own status updates).
-	// TODO(metacontroller#26): Use ObservedGeneration instead of this.
```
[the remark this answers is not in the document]
crimsonfaith91: done

controller/composite/controller.go
```diff
@@ -240,23 +239,14 @@ func (pc *parentController) enqueueParentObject(obj interface{}) {
 }
 
 func (pc *parentController) updateParentObject(old, cur interface{}) {
-	// Ignore updates if the spec hasn't changed (e.g. our own status updates).
+	// Ignore updates where the ResourceVersion changed (not a resync)
+	// but the spec hasn't changed (e.g. our own status updates).
```
crimsonfaith91: but the `generation`

examples/bluegreen/bluegreen-controller.yaml
```diff
@@ -12,6 +12,8 @@ spec:
     kind: BlueGreenDeployment
     shortNames:
     - bgd
+  subresources:
```
crimsonfaith91: Is this necessary? It looks like BGD test passes even without status subresource.

dynamic/discovery/discovery.go
```diff
@@ -89,6 +89,23 @@ func (rm *ResourceMap) GetKind(apiVersion, kind string) (result *APIResource) {
...
+	rm.mutex.RLock()
+	defer rm.mutex.RUnlock()
+
+	gv, ok := rm.groupVersions[parentResource.APIVersion]
+	if !ok {
+		return false
+	}
+
+	for key, _ := range gv.subresources {
+		if key == parentResource.Name+"/"+"status" {
```
crimsonfaith91: For example, the key is "catsets/status".

Gopkg.lock
```diff
@@ -2,233 +2,184 @@
 
 
 [[projects]]
-  digest = "1:3dcf9df58e425f7ace052d832e683930533be49fcde3b6d66cebb2ed5a97bb3f"
```
enisoc: I think these lines might be getting removed because you have an old version of dep. Can you try updating to the latest and running `dep ensure` again?

controller/common/manage_children.go
```diff
@@ -52,6 +52,9 @@ func ApplyUpdate(orig, update *unstructured.Unstructured) (*unstructured.Unstruc
 	if err := revertObjectMetaSystemFields(newObj, orig); err != nil {
 		return nil, fmt.Errorf("failed to revert ObjectMeta system fields: %v", err)
 	}
+	if err := revertStatus(newObj, orig); err != nil {
```
enisoc: Add a comment above this explaining that we revert status because we don't currently support a parent changing the status of its children, so we need to make sure no diffs on children involve status.

controller/common/manage_children.go
```diff
@@ -95,6 +98,26 @@ func revertObjectMetaSystemFields(newObj, orig *unstructured.Unstructured) error
 	return nil
 }
 
+// revertStatus overwrites the .status (i.e. all status fields) in newObj to match
+// what it was in orig by resetting .status directly without peeking into each field.
+func revertStatus(newObj, orig *unstructured.Unstructured) error {
```
enisoc: Optional suggestion: You could change `revertStatus` to `revertField` and add a `fieldPath ...string` parameter at the end. Then this becomes a generic function for reverting anything, and you can call it from inside the `revertObjectMetaSystemFields` function to avoid duplication.

controller/composite/controller.go
enisoc: It's more accurate to leave it as "spec hasn't changed". We can't always rely on `metadata.generation` to know if the spec changed, because not all resources have generation enabled.

controller/composite/controller.go
```diff
@@ -241,13 +241,12 @@ func (pc *parentController) enqueueParentObject(obj interface{}) {
 func (pc *parentController) updateParentObject(old, cur interface{}) {
 	// Ignore updates where the ResourceVersion changed (not a resync)
 	// but the spec hasn't changed (e.g. our own status updates).
-	// TODO(metacontroller#26): Use ObservedGeneration instead of this.
 	oldParent := old.(*unstructured.Unstructured)
 	curParent := cur.(*unstructured.Unstructured)
 	if curParent.GetResourceVersion() != oldParent.GetResourceVersion() {
-		oldSpec := k8s.GetNestedField(oldParent.UnstructuredContent(), "spec")
-		curSpec := k8s.GetNestedField(curParent.UnstructuredContent(), "spec")
-		if reflect.DeepEqual(oldSpec, curSpec) {
```
enisoc: As mentioned above, I still recommend leaving this function unchanged for now. Changing this is going to be more complicated that what you have now (we need to fall back to DeepEqual unless we know generation is being incremented) and it's not even worth it because it'll become less reliable (not every built-in resource that has /status increments generation properly).

dynamic/clientset/clientset.go
```diff
@@ -175,3 +175,32 @@ func (rc *ResourceClient) RemoveFinalizer(orig *unstructured.Unstructured, name
 		return true
 	})
 }
+
+// AtomicStatusUpdate is similar to AtomicUpdate, except that it updates status.
+func (rc *ResourceClient) AtomicStatusUpdate(resourceMap *dynamicdiscovery.ResourceMap, parentResource *dynamicdiscovery.APIResource, orig *unstructured.Unstructured, update func(obj *unstructured.Unstructured) bool) (result *unstructured.Unstructured, err error) {
```
enisoc: We shouldn't need to add `resourceMap` or `parentResource`. This should have the same arguments as `AtomicUpdate`.

dynamic/clientset/clientset.go
```diff
@@ -175,3 +175,32 @@ func (rc *ResourceClient) RemoveFinalizer(orig *unstructured.Unstructured, name
...
+			// The original object was deleted and replaced with a new one.
+			return apierrors.NewNotFound(rc.GroupResource(), name)
+		}
+		if changed := update(current); !changed {
+			// There's nothing to do.
+			result = current
+			return nil
+		}
+
+		if resourceMap.HasSubresource(parentResource, "status") {
```
enisoc: This should be encapsulated into `APIResource` so that you can write it as:

```go
if rc.HasSubresource("status") {
```

That way we don't require the caller to pass in extra arguments. It probably requires storing a list of subresources in the APIResource struct.

dynamic/clientset/clientset.go
```diff
@@ -175,3 +175,32 @@ func (rc *ResourceClient) RemoveFinalizer(orig *unstructured.Unstructured, name
 		return true
 	})
 }
+
+// AtomicStatusUpdate is similar to AtomicUpdate, except that it updates status.
+func (rc *ResourceClient) AtomicStatusUpdate(resourceMap *dynamicdiscovery.ResourceMap, parentResource *dynamicdiscovery.APIResource, orig *unstructured.Unstructured, update func(obj *unstructured.Unstructured) bool) (result *unstructured.Unstructured, err error) {
+	name := orig.GetName()
+
+	err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		current, err := rc.Get(name, metav1.GetOptions{})
```
enisoc: Add a TODO above this line that ideally we would call `GetStatus` (if it HasSubresource) to respect subresource RBAC rules, but the dynamic client doesn't offer that yet.

dynamic/discovery/discovery.go
```diff
@@ -89,6 +89,23 @@ func (rm *ResourceMap) GetKind(apiVersion, kind string) (result *APIResource) {
 	return gv.kinds[kind]
 }
 
+func (rm *ResourceMap) HasSubresource(parentResource *APIResource, subresourceKey string) bool {
```
enisoc: As mentioned above, I would prefer that HasSubresource is a method on `APIResource`, not on `ResourceMap`. That way, APIResource includes everything you need to know, and you don't have to go back to the map to ask more questions if you already have the APIResource (which is already available inside every ResourceClient).

dynamic/discovery/discovery.go
```diff
@@ -89,6 +89,23 @@ func (rm *ResourceMap) GetKind(apiVersion, kind string) (result *APIResource) {
...
+	rm.mutex.RLock()
```
enisoc: Another benefit of denormalizing the knowledge of whether /status exists into every APIResource object is that we won't have to take a lock to read it.

dynamic/discovery/discovery.go
```diff
@@ -131,6 +149,8 @@ func (rm *ResourceMap) refresh() {
 			// Make sure we don't choose a subresource like "pods/status".
 			if !strings.ContainsRune(apiResource.Name, '/') {
 				gve.kinds[apiResource.Kind] = apiResource
+			} else {
+				gve.subresources[apiResource.Name] = apiResource
```
enisoc: After you store all these subresources, do another pass over all subresources to denormalize the info into the corresponding main resource, so that the main APIResource contains a list of subresources and a `HasSubresource` method to query it.

dynamic/discovery/discovery.go
```diff
@@ -131,6 +149,8 @@ func (rm *ResourceMap) refresh() {
 			// Make sure we don't choose a subresource like "pods/status".
 			if !strings.ContainsRune(apiResource.Name, '/') {
 				gve.kinds[apiResource.Kind] = apiResource
+			} else {
```
enisoc: Invert this if/else to get rid of the double negative, now that both branches have code.

Also update the comment about not choosing subresources to clarify, since we now do care about subresources. It should be something like:

```go
// Remember which resources are subresources, and map the kind to the main resource.
```

examples/bluegreen/bluegreen-controller.yaml
enisoc: Yes. The test only passes because we failed to test that status works. If you have time, you could add some checks in `test.sh` to verify that the BGD status is updated correctly. Right now, it's only looking at the underlying RS statuses.

controller/composite/controller.go
crimsonfaith91: I recalled a suggestion of changing this to `.generation` or `.observedGeneration`. Does it mean that the previous suggestion does not apply anymore?

dynamic/discovery/discovery.go
crimsonfaith91: @enisoc 
Another alternative is to denormalize the info on the spot without introducing the another pass. As `gve` maps are of `map[string]*APIResource` type, the change will be reflected immediately (the `APIResource` is a pointer), I think the another pass should be unnecessary.

Do you have any reason for introducing the another pass (e.g., better code readability)?

controller/common/manage_children.go
crimsonfaith91: Yes, but this will remove your comments in `revertObjectMetaSystemFields` function. I will introduce the `revertField` function.

enisoc: Since `revertObjectMetaSystemFields` has more descriptive comments, you can copy those over to the new `revertField`. I believe they still apply.

controller/composite/controller.go
enisoc: The [original suggestion](https://github.com/GoogleCloudPlatform/metacontroller/pull/102/files#r221055866) was a two-parter: (1) this should be generation instead of observedGeneration, and (2) there's more you'll need in addition to this change, and it will become less reliable, so let's not change anything here right now. Sorry for the confusing structure. I should have started with the final recommendation instead of burying it in the middle of a long comment.

dynamic/discovery/discovery.go
enisoc: What we want is to add some data to the APIResource for "foo" while we are processing the entry for "foo/status". The second pass is necessary to make sure that "foo" has already been processed when we look for it to add the subresource data. Even if it's technically the case that the discovery API always returns "foo" before "foo/status" in the list (which I have no reason to suspect anyway), it would be brittle to depend on that ordering.

controller/common/manage_children.go
crimsonfaith91: You are right, the comments still apply. Thanks!

controller/composite/controller.go
crimsonfaith91: Noted, I will revert this function.

dynamic/discovery/discovery.go
crimsonfaith91: I think we should store all subresources, not just `/status`. This fits better with `HasSubresource("someSubresource")` function.

dynamic/discovery/discovery.go
```diff
@@ -18,6 +18,7 @@ package discovery
 
 import (
 	"fmt"
+	"reflect"
```
crimsonfaith91: this is unnecessary, will remove later

dynamic/discovery/discovery.go
```diff
@@ -125,14 +136,39 @@ func (rm *ResourceMap) refresh() {
...
+				// Remember which resources are subresources, and map the kind to the main resource.
+				gve.subresources[apiResource.Name] = apiResource
+			} else {
+				// Make sure we don't choose a subresource like "pods/status".
+				gve.kinds[apiResource.Kind] = apiResource
+			}
+		}
+
+		// Group all subresources for a resource.
+		for i := range group.APIResources {
```
enisoc: You don't have to loop over all resources again. We already kept a map of only the ones that are subresources. In the second pass, you can just loop over all the subresources and insert them where they belong by looking up the main resource (if it exists). That would be O(n) instead of the current code which is O(n^2), and it also avoids allocating a new APIResource by reusing the one we already put in the map.

crimsonfaith91: Yes, your proposed way works better. This O(N^2) is to ensure this is something similar to what you intend for. I spent some time looking for O(N^1), but couldn't find any for recording all subresources for a resource, so I think O(N) is the best solution.

dynamic/discovery/discovery.go
```diff
@@ -139,34 +135,24 @@ func (rm *ResourceMap) refresh() {
...
-					split := strings.Split(apiSubresource.Name, "/")
-					resourceName := split[0]
-					subresourceKey := split[1]
-					if apiResource.Name == resourceName {
-						apiResource.SubresourceMap[subresourceKey] = true
-					}
-				}
-				gve.resources[apiResource.Name] = apiResource
-				gve.kinds[apiResource.Kind] = apiResource
+		for apiSubresourceName, _ := range gve.subresources {
```
crimsonfaith91: @enisoc All tests passed now. Couldn't we achieve the same by adding the logic inside `if strings.ContainsRune(apiResource.Name, '/') {...}`?

dynamic/discovery/discovery.go
```diff
@@ -139,34 +135,24 @@ func (rm *ResourceMap) refresh() {
...
+			arr := strings.Split(apiSubresourceName, "/")
+			apiResourceName := arr[0]
+			subresourceKey := arr[1]
+			apiResource := gve.resources[apiResourceName]
+			if len(apiResource.subresourceMap) == 0 {
```
crimsonfaith91: This looks awkward, but it saves unnecessary initialization of `.subresourceMap` for subresources by not initializing the field for all resources upfront in previous loop.

dynamic/discovery/discovery.go
```diff
@@ -125,14 +132,29 @@ func (rm *ResourceMap) refresh() {
 				apiResource.Version = gv.Version
 			}
 			gve.resources[apiResource.Name] = apiResource
-			// Remember how to map back from Kind to resource.
+			// Remember how to map back from Kind/subresource to resource.
```
enisoc: I meant for the replacement comment to go here.

```suggestion
			// Remember which resources are subresources, and map the kind to the main resource.
```

dynamic/discovery/discovery.go
```diff
@@ -125,14 +132,29 @@ func (rm *ResourceMap) refresh() {
...
 			// This is different from what RESTMapper provides because we already know
 			// the full GroupVersionKind and just need the resource name.
-			// Make sure we don't choose a subresource like "pods/status".
-			if !strings.ContainsRune(apiResource.Name, '/') {
+			// Remember which resources are subresources, and map the kind to the main resource.
```
enisoc:
```suggestion
```

dynamic/discovery/discovery.go
enisoc: I think the intent is more clear if we compare with `nil`. We don't really care whether there are any items. We care if the map has been allocated or not.

```suggestion
			if apiResource.subresourceMap == nil {
```

dynamic/discovery/discovery.go
```diff
@@ -125,14 +132,29 @@ func (rm *ResourceMap) refresh() {
...
 				gve.kinds[apiResource.Kind] = apiResource
 			}
 		}
+
+		// Group all subresources for a resource.
+		for apiSubresourceName, _ := range gve.subresources {
+			arr := strings.Split(apiSubresourceName, "/")
+			apiResourceName := arr[0]
+			subresourceKey := arr[1]
+			apiResource := gve.resources[apiResourceName]
```
enisoc: Just in case the discovery result is something unexpected, let's check here if `apiResource` is `nil` and skip it if so. We don't want to panic in that case.

controller/composite/controller.go
```diff
@@ -562,15 +561,15 @@ func (pc *parentController) claimChildren(parent *unstructured.Unstructured) (co
...
-	return pc.parentClient.Namespace(parent.GetNamespace()).AtomicUpdate(parent, func(obj *unstructured.Unstructured) bool {
+	return pc.parentClient.Namespace(parent.GetNamespace()).AtomicStatusUpdate(parent, func(obj *unstructured.Unstructured) bool {
 		oldStatus := k8s.GetNestedField(obj.UnstructuredContent(), "status")
 		if reflect.DeepEqual(oldStatus, status) {
 			// Nothing to do.
 			return false
 		}
+
 		k8s.SetNestedField(obj.UnstructuredContent(), status, "status")
+		k8s.SetNestedField(obj.UnstructuredContent(), parent.GetGeneration(), "status", "observedGeneration")
```
enisoc: Sorry for not noticing earlier, but we should add a check in at least one of the tests that status.observedGeneration is getting set correctly.

crimsonfaith91:
@enisoc I have updated my `dep` version and ran `dep ensure`. Let me try it again.

dynamic/discovery/discovery.go
crimsonfaith91: Good point.
```
if apiResource == nil {
    continue
}
```

controller/composite/controller.go
crimsonfaith91: In the long run, I suggest adding integration tests.

crimsonfaith91:
@enisoc Ready for final round of review, thanks!

PR #105, opened 2018-10-01
Clarify docs for generateSelector in CompositeController.

### LINKED ISSUE #104: Allow CompositeController to use owner references instead of a selector
I have a use case for controlling a resource whose definition isn't owned by me and has no `spec.selector` field. There will only ever be one or two child resources, so a selector isn't really necessary. DecoratorController won't work because it doesn't allow me to update `status`. CompositeController won't work because it requires a `spec.selector` field.

What I want is a hybrid of the two strategies: The capabilities of CompositeController with the owner referencing strategy of DecoratorController.

Would it be possible to extend CompositeController with an option to allow using owner references for tracking child objects instead of a selector?

enisoc:
@grantr Does this rewrite answer your questions from #104 more clearly?

Fixes #104

### DIFF
--- a/docs/_api/compositecontroller.md
+++ b/docs/_api/compositecontroller.md
@@ -214,23 +214,39 @@ a no-op (no CRUD operations needed to achieve desired state).
 
 ## Generate Selector
 
-Usually, each parent object managed by a CompositeController has its own
+Usually, each parent object managed by a CompositeController must have its own
 user-specified [label selector](#label-selector), just like each
-Deployment has its own label selector.
+Deployment has its own label selector in `spec.selector`.
+However, sometimes it makes more sense to let the user of your API pretend there
+are no labels or label selectors.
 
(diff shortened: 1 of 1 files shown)


docs/_api/compositecontroller.md
```diff
@@ -214,23 +214,39 @@ a no-op (no CRUD operations needed to achieve desired state).
...
+On the other hand, a ReplicaSet assumes all Pods that match its selector are
+interchangeable, so it would be happy to have one less replica it has to create.
+
+If you set `spec.generateSelector` to `true` in your CompositeController
+definition, Metacontroller will do the following:
+
+* When creating children for you, Metacontroller will automatically add a label
+  that points to the parent object's unique ID (`metadata.uid`).
+* Metacontroller will *not* expect each parent object to contain a
+  `spec.selector`, and will ignore the value even if one of set.
```
grantr: typo nit: `even if one is set`

PR #112, opened 2018-10-23
CompositeController: Add label and selector safety checks

### LINKED ISSUE #24: CompositeController should validate that child objects match the parent selector.
We should fail creation/update of child objects that don't match, because otherwise they'll be immediately orphaned.

There might be legitimate use cases for a controller that intentionally orphans objects, but I don't think it should be allowed by default when using CompositeController, since that's contrary to the intended pattern: the parent should represent a "composite" object that's made up of child objects.

### LINKED ISSUE #25: CompositeController shouldn't allow empty parent selector.
A key feature of CompositeController is that it tracks which objects of a given type are yours, and hides from you any objects of that type that you don't own. This mechanism is defeated if the parent selector is empty, since there's no way to distinguish things you want from things you don't.

There might be legitimate use cases for a controller that selects everything of a given type, but I don't think it should be allowed by default in CompositeController, since more often than not an empty selector is unintentional. For example, this can happen if the parent object does not have the expected, duck-typed `spec.selector` (either because it's missing, or it's the old-style selector like in RC or Service).

### LINKED ISSUE #109: Invalid labels map is dropped silently
I need to attach some simple string metadata to Pods created by my custom CompositeController. I've tried to include such metadata in pod specs returned in the controller's desired list of children as
- `children[].metadata.<my_metadata>`
- `children[].metadata.annotations.<my_metadata>`
- `children[].metadata.labels.<my_metadata>`

All of these items in the JSON object are accepted but ignored by Metacontroller. When I inspect the running Pods using `kubectl get pod <my_pod> -oyaml`, the metadata simply isn't there.

Is this by design, or is there something I can do differently?

enisoc:
Fixes #24
Fixes #25
Fixes #109

### DIFF
--- a/controller/composite/controller.go
+++ b/controller/composite/controller.go
@@ -490,6 +490,11 @@ func (pc *parentController) makeSelector(parent *unstructured.Unstructured, extr
 		if err := k8s.GetNestedFieldInto(labelSelector, parent.UnstructuredContent(), "spec", "selector"); err != nil {
 			return nil, fmt.Errorf("can't get label selector from %v %v/%v", pc.parentResource.Kind, parent.GetNamespace(), parent.GetName())
 		}
+		// An empty selector doesn't make sense for a CompositeController parent.
+		// This is likely user error, and could be dangerous (selecting everything).
+		if len(labelSelector.MatchLabels) == 0 && len(labelSelector.MatchExpressions) == 0 {
+			return nil, fmt.Errorf(".spec.selector must have either matchLabels, matchExpressions, or both")
+		}
 	}
 
 	for key, value := range extraMatchLabels {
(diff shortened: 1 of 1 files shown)


controller/composite/controller.go
```diff
@@ -434,20 +434,34 @@ func (pc *parentController) syncParentObject(parent *unstructured.Unstructured)
...
+	}
+	for _, group := range desiredChildren {
+		for _, obj := range group {
+			// We don't use GetLabels() because that swallows conversion errors.
+			objLabels, _, err := unstructured.NestedStringMap(obj.UnstructuredContent(), "metadata", "labels")
+			if err != nil {
+				return fmt.Errorf("invalid labels on desired child %v %v/%v: %v", obj.GetKind(), obj.GetNamespace(), obj.GetName(), err)
+			}
+			// If selector generation is enabled, add the controller-uid label to all
+			// desired children.
```
crimsonfaith91: Suggestion: add more details on why `controller-uid` label is needed when selector generation is enabled.

enisoc: Done.

PR #121, opened 2018-11-07
Split dev kustomization into base + dev overlay

vreon:
This allows Kustomize users to refer to Metacontroller's base
kustomization by URL, either for installation:

    $ kustomize build github.com/GoogleCloudPlatform/metacontroller?ref=v0.4.0 | kubectl apply -f -

Or as a base in their own overlays:

```yaml
bases:
- github.com/GoogleCloudPlatform/metacontroller?ref=v0.4.0
```

(P.S. Thanks for Metacontroller!!)


enisoc:
Nice idea. Thanks!

PR #122, opened 2018-11-16
bugfix for finalizers to run

### LINKED ISSUE #125: DeletionTimestamp being added to parent resource doesn't trigger a resync
Hi, I've been looking at using composite controller with an empty list of children to control an external service via my own CRD. Everything works fine with regard to creating resources and updating them and I don't currently need to put changes from the external system back into the resource. 

The issue I have is that metacontroller doesn't seem to notice when a deletionTimestamp is placed in the metadata of a parent resource. So when I try to delete the custom resource, the deletionTimestamp is created by k8s but metacontroller doesn't react with a request to my controller. If I modify the resource in some other way though (changing something in its spec), then metacontroller sees the change, notices the deletionTimestamp and deletes the resource after the finalize request to my controller. It's just a hack project so I can work around it for now via a resync period but it's not ideal.

Is this intended behaviour or even a desired use-case?

bsherrod:
Bugfix for observed behavior where CompositeController finalizers do not run, in kubernetes v1.10.8.

The `updateParentObject()` method ignores updates when the resource version and spec haven't changed.  However, `kubectl delete` , when finalizers are present, will add `deletionTimestamp`, but will not modify the resource version or the spec.   So finalizers do not run.

This PR changes `updateParentObject()` to always process updates if they contain a deletion timestamp, so that the finalizers run.

Fixes #125

### DIFF
--- a/controller/composite/controller.go
+++ b/controller/composite/controller.go
@@ -243,7 +243,9 @@ func (pc *parentController) updateParentObject(old, cur interface{}) {
 	// but the spec hasn't changed (e.g. our own status updates).
 	oldParent := old.(*unstructured.Unstructured)
 	curParent := cur.(*unstructured.Unstructured)
-	if curParent.GetResourceVersion() != oldParent.GetResourceVersion() {
+	if curParent.GetDeletionTimestamp() == nil &&
+		oldParent.GetDeletionTimestamp() == nil &&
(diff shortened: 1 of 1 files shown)


bsherrod:
A similar change may be necessary with DecoratorControllers, but I haven't checked (not using them).

bsherrod:
i signed it!

bsherrod:
@enisoc PTAL

PR #124, opened 2018-12-17
Fix deletion of namespaced children

### LINKED ISSUE #127: A cluster-scoped controller cannot update or delete namespaced resources
I'm writing a cluster-scoped `CompositeController` that sometimes needs to create `Pods`, which are of course namespaced.

I don't think the support for this in Metacontroller is working. When my controller omits a Pod it had created from the `children` it returns, metacontroller can't delete it, and logs an error on every reconcile:

```
can't reconcile children for <the custom resource>: can't delete Pod <ns>/<name>: the server could not find the requested resource
```

I'm guessing that `ManageChildren`'s `deleteChildren` helper [chokes](https://github.com/GoogleCloudPlatform/metacontroller/blob/master/controller/common/manage_children.go#L173) on the `namespace/name` keys Metacontroller generates for namespace-scoped children with a cluster-scoped parent:

```go
for name, obj := range observed {
	if desired == nil || desired[name] == nil {
		err := client.Namespace(obj.GetNamespace()).Delete(name, …)
	}
}
```

If `name` here is really `"namespace/name"`, that's not going to work. It looks like `updateChildren` is affected as well.

mgruener:
Bugfix for the deletion of namespaced child resources.

When a cluster scoped parent resource manages namespace scoped child resources, the `Delete()` method gets passed the namespaced resource name in the format `{.metadata.namespace}/{.metadata.name}` instead of just the resource name.

This results in errors of the form

```
[[ can't delete <resource> <namespace>/<resource name>: the server could not find the requested resource] the server could not find the requested resource]
```

This PR ensures that the `Delete()` functions gets passed the resource name without the namespace part.

Fixes #127

### DIFF
--- a/controller/common/manage_children.go
+++ b/controller/common/manage_children.go
@@ -166,11 +166,15 @@ func deleteChildren(client *dynamicclientset.ResourceClient, parent *unstructure
 			// Skip objects that are already pending deletion.
 			continue
 		}
+		ns := obj.GetNamespace()
+		if ns == "" {
+			ns = parent.GetNamespace()
+		}
 		if desired == nil || desired[name] == nil {
 			// This observed object wasn't listed as desired.
 			glog.Infof("%v: deleting %v", describeObject(parent), describeObject(obj))
 			uid := obj.GetUID()
-			err := client.Namespace(obj.GetNamespace()).Delete(name, &metav1.DeleteOptions{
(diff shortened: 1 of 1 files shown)


enisoc:
I went ahead and merged #128 since it was already what you'd get after addressing review comments on this PR. Thanks for the fix though and sorry for the slow response. I was out on leave but back now and trying to catch up on the backlog.

PR #128, opened 2019-01-12
Fix deleting a ns-scoped child of a cluster-scoped controller

### LINKED ISSUE #127: A cluster-scoped controller cannot update or delete namespaced resources

silverlyra:
Fixes #127.

I tried a build with this change in my cluster, and it was able to delete a pod created on behalf of a cluster-scoped composite controller.

### DIFF
--- a/controller/common/manage_children.go
+++ b/controller/common/manage_children.go
@@ -170,7 +170,7 @@ func deleteChildren(client *dynamicclientset.ResourceClient, parent *unstructure
 			// This observed object wasn't listed as desired.
 			glog.Infof("%v: deleting %v", describeObject(parent), describeObject(obj))
 			uid := obj.GetUID()
-			err := client.Namespace(obj.GetNamespace()).Delete(name, &metav1.DeleteOptions{
+			err := client.Namespace(obj.GetNamespace()).Delete(obj.GetName(), &metav1.DeleteOptions{
 				Preconditions: &metav1.Preconditions{UID: &uid},
 			})
 			if err != nil {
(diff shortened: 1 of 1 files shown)


mgruener:
This is the same fix as #124 or am I missing something?

silverlyra:
@mgruener – ah, I hadn’t seen your PR when I wrote mine. They’re definitely fixing the same bug.

in #124, I don’t understand when it’d be necessary to use `parent.GetNamespace()`. Was `obj.GetNamespace() == ""` a case you ran into?

enisoc:
Actually, I take that back. After reviewing #124, it turns out this PR is already doing what you'd get after addressing the review comments on the other one, so I'll just merge this.

PR #129, opened 2019-01-15
Delete namespace resource from kustomize

mgoodness:
Would like to use these manifests as a base for kustomize, but with my own namespace and not have the `metacontroller` namespace created. Shouldn't be necessary anyway, as most (all?) other tools will create the namespace automatically for the other resources.

### DIFF
--- a/controller/common/manage_children.go
+++ b/controller/common/manage_children.go
@@ -170,7 +170,7 @@ func deleteChildren(client *dynamicclientset.ResourceClient, parent *unstructure
 			// This observed object wasn't listed as desired.
 			glog.Infof("%v: deleting %v", describeObject(parent), describeObject(obj))
 			uid := obj.GetUID()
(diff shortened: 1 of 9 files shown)


enisoc:
When I try `kubectl apply -f manifests/metacontroller-rbac.yaml` with your patch and kubectl v1.11.3, it fails to create the ServiceAccount with error `namespaces "metacontroller" not found`.

Is there another way to make the `metacontroller` namespace optional when used as a kustomize base, but still include it by default when using these manifests directly with `kubectl apply`?

mgoodness:
Oops, you're right. Not sure what I was thinking.

We could pull the namespace out into its own `manifests/namespace.yaml` file and not include that in `kustomization.yaml`. Then one could run `kubectl apply -f manifests/` and get the desired effect.

mgoodness:
Oops again. Closing and will resubmit. Sorry.

PR #133, opened 2019-01-16
Move namespace to own manifest

mgoodness:
Allows `kubectl apply -f manifests/` to create all required resources, including a `metacontroller` Namespace.

However, `kustomization.yaml` does not include that manifest in its `resources`, which lets kustomize handle namespace creation.

Replaces my borked #129.

### DIFF
--- a/manifests/metacontroller-rbac.yaml
+++ b/manifests/metacontroller-rbac.yaml
@@ -1,9 +1,4 @@
 apiVersion: v1
-kind: Namespace
-metadata:
-  name: metacontroller
----
-apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: metacontroller
--- a/manifests/namespace.yaml
+++ b/manifests/namespace.yaml
@@ -1,0 +1,4 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: metacontroller


enisoc:
To get `kubectl apply -f manifests/` to work, I had to rename `namespace.yaml` to `kubernetes-namespace.yaml` so it comes before `kubernetes.yaml` in the sort order.

I also tried with the latest Skaffold 0.20.0, and I'm not seeing it create the namespace automatically with either `skaffold run` or `skaffold run -n metacontroller`. Is it a feature of kustomize that it creates the namespace for you implicitly? It seems uncommon for tools to do that.

mgoodness:
> To get kubectl apply -f manifests/ to work, I had to rename namespace.yaml to kubernetes-namespace.yaml so it comes before kubernetes.yaml in the sort order.

I can prefix the manifests so kubectl loads them in the correct order.

> Is it a feature of kustomize that it creates the namespace for you implicitly?

No, kustomize doesn't create namespaces implicitly. My original comment was misworded.

That said, I think the common pattern so far is to not include a namespace manifest in the base layer. End users then have a couple of options in their overlays.

- kustomize creates a namespace:
```
bases:
- github.com/GoogleCloudPlatform/metacontroller

resources:
- namespace.yaml
```

- namespace is created through some other process then "consumed" by kustomize:
```
bases:
- github.com/GoogleCloudPlatform/metacontroller

namespace: my-namespace
```

The former seems like a good fit for tools like skaffold, while the latter is what I'm trying to do. Should I create a second `kustomization.yaml` for skaffold?

enisoc:
Thanks for clarifying. I added some commits after yours that should maintain the intent of your PR while keeping our Skaffold workflow unchanged. Please let me know if you find any problems.

mgoodness:
Perfect. Thanks!

PR #136, opened 2019-01-23
Add integration test framework and sample test.

enisoc:
This adds a framework for integration tests that run against a bare kube-apiserver with no other Kubernetes components (i.e. no kubelets). Unlike the end-to-end tests in `examples/`, these integration tests don't require an actual Kubernetes cluster. This allows efficient testing of Metacontroller behavior at the level of its interaction with Kubernetes API objects, after which the respective implementations of those APIs are assumed to work.

This also includes the first sample test, which exercises a simple CompositeController sync webhook scenario to validate the test framework. Additional tests will be added in separate PRs.

### DIFF
--- a/hack/get-kube-binaries.sh
+++ b/hack/get-kube-binaries.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+
+set -e
+set -u
+
+# This script downloads etcd and Kubernetes binaries that are
+# used as part of the integration test environment,
+# and places them in hack/bin/.
+#
+# The integration test framework expects these binaries to be found in the PATH.
+
+# This is the kube-apiserver version to test against.
+KUBE_VERSION="${KUBE_VERSION:-v1.11.3}"
+KUBERNETES_RELEASE_URL="${KUBERNETES_RELEASE_URL:-https://dl.k8s.io}"
+
+# This should be the etcd version downloaded by kubernetes/hack/lib/etcd.sh
+# as of the above Kubernetes version.
+ETCD_VERSION="${ETCD_VERSION:-v3.2.18}"
+
+mkdir -p hack/bin
+cd hack/bin
+
+# Download kubectl.
+rm -f kubectl
+wget "${KUBERNETES_RELEASE_URL}/${KUBE_VERSION}/bin/linux/amd64/kubectl"
+chmod +x kubectl
+
+# Download kube-apiserver.
+rm -f kube-apiserver
+wget "${KUBERNETES_RELEASE_URL}/${KUBE_VERSION}/bin/linux/amd64/kube-apiserver"
+chmod +x kube-apiserver
+
+# Download etcd.
+rm -f etcd
+basename="etcd-${ETCD_VERSION}-linux-amd64"
+filename="${basename}.tar.gz"
+url="https://github.com/coreos/etcd/releases/download/${ETCD_VERSION}/${filename}"
+wget "${url}"
+tar -zxf "${filename}"
+mv "${basename}/etcd" etcd
+rm -rf "${basename}" "${filename}"
--- a/server/server.go
+++ b/server/server.go
@@ -1,0 +1,101 @@
+/*
+Copyright 2019 Google Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package server
+
+import (
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/0xRLG/ocworkqueue"
+	"go.opencensus.io/stats/view"
+
+	"k8s.io/client-go/discovery"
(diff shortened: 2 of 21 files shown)


test/integration/framework/metacontroller.go
```diff
@@ -0,0 +1,69 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	apiextensions "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+
+	"metacontroller.app/apis/metacontroller/v1alpha1"
+)
+
+// CreateCompositeController generates a test CompositeController and installs
+// it in the test API server.
+func (f *Fixture) CreateCompositeController(name, syncHookURL string, parentCRD, childCRD *apiextensions.CustomResourceDefinition) *v1alpha1.CompositeController {
```
crimsonfaith91: Is there any reason we don't return `(*v1alpha1.CompositeController, error)` here? In other words, why `Fatal` is preferred than `error`?

test/integration/framework/main.go
```diff
@@ -0,0 +1,102 @@
...
+
+	dynamicdiscovery "metacontroller.app/dynamic/discovery"
+	"metacontroller.app/server"
+)
+
+var resourceMap *dynamicdiscovery.ResourceMap
+
+// TestMain starts etcd, kube-apiserver, and metacontroller before running tests.
+func TestMain(tests func() int) {
+	var result int
```
crimsonfaith91: Should we assign a non-zero int value before calling `defer`?

test/integration/framework/main.go
```diff
@@ -0,0 +1,102 @@
...
+			break
+		}
+		if time.Since(start) > defaultWaitTimeout {
+			klog.Fatalf("timed out waiting for kube-apiserver to be ready: %v", err)
+		}
+		time.Sleep(time.Second)
+	}
+
+	// Install Metacontroller RBAC.
+	if err := execKubectl("apply", "-f", "../../../manifests/metacontroller-rbac.yaml"); err != nil {
```
crimsonfaith91: Should we set `../../../manifests/` as a global variable? It looks like we may want to add more manifests to the path, and the path will be called more frequently.

test/integration/framework/crd.go
```diff
@@ -0,0 +1,103 @@
...
+
+	client, err := f.dynamic.Resource(APIVersion, plural)
+	if err != nil {
+		f.t.Fatal(err)
+	}
+
+	f.t.Logf("Waiting for %v CRD client List() to succeed...", kind)
+	err = f.Wait(func() (bool, error) {
+		_, err := client.List(metav1.ListOptions{})
+		return err == nil, nil
```
crimsonfaith91: `return err == nil, err`?

test/integration/framework/apiserver.go
```diff
@@ -0,0 +1,115 @@
...
+	"k8s.io/client-go/rest"
+	"k8s.io/klog"
+)
+
+var apiserverURL = ""
+
+const installApiserver = `
+Cannot find kube-apiserver, cannot run integration tests
+
+Please download kube-apiserver and ensure it is somewhere in the PATH.
```
crimsonfaith91: Should we also provide a link to download it?

test/integration/composite/composite_test.go
```diff
@@ -0,0 +1,82 @@
...
+	unstructured.SetNestedStringMap(parent.Object, labels, "spec", "selector", "matchLabels")
+	_, err := parentClient.Namespace(ns).Create(parent)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	t.Logf("Waiting for child object to be created...")
+	err = f.Wait(func() (bool, error) {
+		_, err = childClient.Namespace(ns).Get("test-sync-webhook", metav1.GetOptions{})
+		return err == nil, nil
```
crimsonfaith91: `return err == nil, err` here?

server/server.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func Start(config *rest.Config, discoveryInterval, informerRelist time.Duration) (stop func(), err error) {
+	// Periodically refresh discovery to pick up newly-installed resources.
+	dc := discovery.NewDiscoveryClientForConfigOrDie(config)
+	resources := dynamicdiscovery.NewResourceMap(dc)
+	// We don't care about stopping this cleanly since it has no external effects.
+	resources.Start(discoveryInterval)
+
+	// Create informerfactory for metacontroller api objects.
```
crimsonfaith91: nit: `informer factory`, so to be coherent with others such as `dynamic clientset` for `dynamicclientset`.

crimsonfaith91:
@enisoc Adding a brief documentation about running the tests will be helpful (e.g., `make test`).

crimsonfaith91:
@enisoc Adding a brief documentation about running the tests will be helpful (e.g., make test).

hack/get-kube-binaries.sh
```diff
@@ -0,0 +1,41 @@
...
+set -u
+
+# This script downloads etcd and Kubernetes binaries that are
+# used as part of the integration test environment,
+# and places them in hack/bin/.
+#
+# The integration test framework expects these binaries to be found in the PATH.
+
+# This is the kube-apiserver version to test against.
+KUBE_VERSION="${KUBE_VERSION:-v1.11.3}"
```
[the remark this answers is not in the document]
crimsonfaith91: Good point! Learnt a new thing today. :)

test/integration/framework/metacontroller.go
crimsonfaith91: I see. Thanks for the explanation!

PR #137, opened 2019-01-25
feat: Add aggregated roles

mikebryant:
From https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles

### DIFF
--- a/manifests/metacontroller-rbac.yaml
+++ b/manifests/metacontroller-rbac.yaml
@@ -33,3 +33,45 @@ roleRef:
   kind: ClusterRole
   name: metacontroller
   apiGroup: rbac.authorization.k8s.io
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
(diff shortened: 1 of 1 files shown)


enisoc:
I haven't used aggregated roles before, but according to the doc link, the admin/edit/view roles are intended for granting access within a namespace. Since most Metacontroller APIs are currently cluster-scoped, is there still some benefit to adding this aggregation?

mikebryant:
Within a namespace, being able to view/edit `controllerrevisions` can be useful, for people trying to work out what's happening with their resource.

For our users, we also use a `ClusterRoleBinding` to grant `view` access to all namespaces and cluster-scoped objects - having the aggregated role there is convenient. (We have some users who are developing controllers based on metacontroller)

It certainly isn't strictly necessary to have this in the public repo, we've already applied this aggregated role to our clusters, but I thought it might be useful for others too.

enisoc:
That makes sense. Thanks!

PR #141, opened 2019-01-29
Update FAQ to reflect latest features.

enisoc:


docs/faq.md
```diff
@@ -56,28 +56,31 @@ APIs that already exist as Kubernetes resources, such as:
...
-[resync period]: /api/compositecontroller/#resync-period
-[issues]: {{ site.repo_url }}/issues
+Metacontroller is not a good fit when you need to examine a large number of
+objects to answer a single hook request.
+For example, if you need to be sent a list of all Pods or all Nodes in order to
+decide on your desired state, we'd have to call your hook with the full list of
+all Pods or Nodes any time any one of them changed.
+However, it might be a good fit if your desired behavior can be naturally
+broken down into per-Pod or per-Node tasks, since then we only need to send your
+hook the objects that have changed.
```
crimsonfaith91: ... your hook _to_ the objects that have changed ...?

docs/faq.md
```diff
@@ -176,6 +179,30 @@ Metacontroller will take care of [reconciling your desired state][reconciling].
 [watches]: /features/#declarative-watches
 [reconciling]: /features/#declarative-reconcilitation
 
+### Can I call external APIs from my hook?
+
+Yes. Your hook code can do whatever it wants as part of computing a response to
+a Metacontroller hook request, including calling external APIs.
+
+The main thing to be careful of is to avoid synchronously waiting for
```
crimsonfaith91: ... to avoid _synchronous waits_ for ... ?

docs/faq.md
enisoc: It's the other way around. We send the objects to the hook; we don't send the hook to the objects.

I can see how the wording is ambiguous though. Do you think it would be easier to read if it were reworded like this?

> ... we only need to send the objects that have changed to your hook.

docs/faq.md
enisoc: Do you find the original wording confusing? As long as the intended meaning is clear, I'd rather keep "waiting" since it sounds more colloquial. Treating "waits" as a noun in this context feels more like technical jargon to me.

crimsonfaith91: I found the original wording to be slightly confusing, but I understand what the sentence means, so it is fine to keep the original wording. :)

docs/faq.md
crimsonfaith91: Oh, misunderstood the description. How about `... we only need to send the changed objects to your hook.`?

PR #142, opened 2019-01-31
docs: Start defining the related objects API

mikebryant:
Part of #98

### DIFF
--- a/docs/_api/compositecontroller.md
+++ b/docs/_api/compositecontroller.md
@@ -258,7 +258,6 @@ Within the CompositeController `spec`, the `hooks` field has the following subfi
 | ----- | ----------- |
 | [`sync`](#sync-hook) | Specifies how to call your sync hook, if any. |
 | [`finalize`](#finalize-hook) | Specifies how to call your finalize hook, if any. |
-| [`related`](#related-hook) | Specifies how to call your related hook, if any. |
 
 Each field of `hooks` contains [subfields][hook] that specify how to invoke
 that hook, such as by sending a request to a [webhook][].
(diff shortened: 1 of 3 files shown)


enisoc:
This is a great start! Out of curiosity, are you also interested in working on code for this after the design is agreed upon? :)

mikebryant:
Kinda, though unsure if I'll be able to commit enough time. This is a personal project, for now.

mikebryant:
One open question for me is, what format should the `related` field be in the sync hook - how do we tell the hook about related objects. Is having it in the same form as the children a good idea?

docs/_api/compositecontroller.md
```diff
@@ -497,3 +498,63 @@ a chance to recheck the external state without holding up a slot in the work
...
+| `labelSelectors` | A list of JSON objects representing all the desired label selectors for this parent object. |
+
+The `labelSelectors` field should contain a flat list of objects,
+not an associative array.
+
+Each labelSelector object should be a JSON object with the following fields:
+| Field | Description |
+| ----- | ----------- |
+| `apiVersion` | The API Version |
+| `kind` | The API Kind |
```
enisoc: In the CompositeController spec, we ask for "resource" instead of "kind". We should probably match that for now, although in the future maybe we should allow the user to give us either resource or kind, since we can translate between them via discovery API anyway.

docs/_api/compositecontroller.md
```diff
@@ -497,3 +498,63 @@ a chance to recheck the external state without holding up a slot in the work
...
+| `controller` | The whole CompositeController object, like what you might get from `kubectl get compositecontroller <name> -o json`. |
+| `parent` | The parent object, like what you might get from `kubectl get <parent-resource> <parent-name> -o json`. |
+
+#### Related Hook Response
+
+The body of your response should be a JSON object with the following fields:
+
+| Field | Description |
+| ----- | ----------- |
+| `labelSelectors` | A list of JSON objects representing all the desired label selectors for this parent object. |
```
enisoc: I would prefer to invert this, so that `labelSelector` is one optional field within a larger object similar to [ResourceRule](https://sourcegraph.com/github.com/GoogleCloudPlatform/metacontroller@353b7d6e7168d478e5fe0aecc716ced192bd92d7/-/blob/apis/metacontroller/v1alpha1/types.go#L47-50). I think that would help make it feel similar to the `childResources` in CompositeController, except this list you get to decide at runtime, whereas that one has to be the same for all parents.

So the hook response could have a field like `relatedResources` which is a list of RelatedResourceRules, which anonymously embed ResourceRule just like [XyzParentResourceRule](https://sourcegraph.com/github.com/GoogleCloudPlatform/metacontroller@353b7d6e7168d478e5fe0aecc716ced192bd92d7/-/blob/apis/metacontroller/v1alpha1/types.go#L52-55), and then also adds `labelSelector` and `namespace`.

mikebryant:
Have moved it to it's own file in `_design`

docs/_api/compositecontroller.md
mikebryant: Sure, have done this

docs/_api/compositecontroller.md
```diff
@@ -258,6 +258,7 @@ Within the CompositeController `spec`, the `hooks` field has the following subfi
 | ----- | ----------- |
 | [`sync`](#sync-hook) | Specifies how to call your sync hook, if any. |
 | [`finalize`](#finalize-hook) | Specifies how to call your finalize hook, if any. |
+| [`related`](#related-hook) | Specifies how to call your related hook, if any. |
```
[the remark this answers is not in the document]
mikebryant: I've renamed the hook to `customize`. It's as good as any for now. I like the thought of potentially using it for more things in future, I've added that

docs/_design/customize-hook.md
```diff
@@ -0,0 +1,67 @@
+### Controller changes
```
enisoc: For some reason this page doesn't show up in the [website preview](https://deploy-preview-142--metacontroller.netlify.com/design/). Maybe we just need to add the front matter at the top of the file:

```
---
title: Customize Hook
---
```

docs/_design/customize-hook.md
```diff
@@ -0,0 +1,67 @@
...
+The `relatedResources` field should contain a flat list of objects,
+not an associative array.
+
+Each resource rule object should be a JSON object with the following fields:
+| Field | Description |
+| ----- | ----------- |
+| `apiVersion` | The API `<group>/<version>` of the parent resource, or just `<version>` for core APIs. (e.g. `v1`, `apps/v1`, `batch/v1`) |
+| `resource`   | The canonical, lowercase, plural name of the parent resource. (e.g. `deployments`, `replicasets`, `statefulsets`) |
+| `labelSelector` | A `v1.LabelSelector` object. |
+| `namespace` | Optional. The Namespace to select in |
```
enisoc: Your example about ConfigMaps and other references made me think: it will probably also be common for controllers to know exactly the name of the object they want. What do you think about adding another optional field at the end here called `names` which is a string list so you can request multiple objects in a namespace with one rule?

mikebryant: yes, this would be useful. Should `labelSelector` be optional too in that case, so you can express one or the other?

docs/_design/customize-hook.md
mikebryant: Fixed :)

docs/_design/customize-hook.md
mikebryant: Or maybe that's implied already, I'm a little hazy on how null vs empty LabelSelectors work.

mikebryant:
I'm pretty happy with this doc, and I'll give prototyping a go

PR #144, opened 2019-02-07
Add basic integration test for DecoratorController.

enisoc:
This is a skeleton for DecoratorController integration tests, just like the one for CompositeController. This also fixes the test framework so integration tests can run in parallel.

### DIFF
--- a/test/integration/framework/main.go
+++ b/test/integration/framework/main.go
@@ -17,6 +17,7 @@ limitations under the License.
 package framework
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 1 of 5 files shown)


test/integration/framework/main.go
```diff
@@ -51,52 +52,55 @@ func getKubectlPath() (string, error) {
...
-			klog.Fatalf("timed out waiting for kube-apiserver to be ready: %v", err)
+			return fmt.Errorf("timed out waiting for kube-apiserver to be ready: %v", kubectlErr)
 		}
 		time.Sleep(time.Second)
 	}
 
 	// Create Metacontroller Namespace.
 	if err := execKubectl("apply", "-f", path.Join(manifestDir, "metacontroller-namespace.yaml")); err != nil {
-		klog.Fatalf("can't install metacontroller namespace: %v", err)
+		return fmt.Errorf("can't install metacontroller namespace: %v", err)
```
crimsonfaith91: nit: `cannot`

crimsonfaith91:
The codes look good to me.
There is a timeout on travis though: `timed out waiting for kube-apiserver to be ready: exit status 1`

crimsonfaith91:
@enisoc Perhaps we can increase [defaultWaitTimeout](https://github.com/GoogleCloudPlatform/metacontroller/blob/master/test/integration/framework/main.go#L81) duration to 1 minute? I understand that integration test setup should be fast, but 10 seconds is a little bit short.

enisoc:
@crimsonfaith91 Yup that did it. The Travis VM is just slower than my workstation. :)

PR #146, opened 2019-02-11
Always delete children with cascading deletion.

### LINKED ISSUE #145: Delete cascade for Job.batch/v1
Delete all pods related to the job after job is deleted from desired children list

enisoc:
For most API types, the server-side default (when the client doesn't specify) is cascading deletion, meaning that the GC will clean up children owned by the object being deleted. For some legacy APIs like Job that went GA before the server-side GC was added, the default is instead orphaning, to maintain backwards compatibility.

Now that server-side cascading deletion (GC) is the norm, users expect it to happen uniformly. Therefore, we should explicitly request cascading deletion whenever we delete children in Metacontroller, in case the server-side default is different for legacy reasons.

Fixes #145

### DIFF
--- a/controller/common/manage_children.go
+++ b/controller/common/manage_children.go
@@ -170,8 +170,12 @@ func deleteChildren(client *dynamicclientset.ResourceClient, parent *unstructure
 			// This observed object wasn't listed as desired.
 			glog.Infof("%v: deleting %v", describeObject(parent), describeObject(obj))
 			uid := obj.GetUID()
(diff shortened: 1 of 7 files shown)


crimsonfaith91:
/lgtm
Should we leave a release note for this change?

enisoc:
Yes, this needs a release note. I write the release notes manually by reading the entire list of commits, so there's no additional action needed besides merging the PR. :)

PR #153, opened 2019-03-06
Add per-object, one-time resync.

### LINKED ISSUE #152: Let the webhook choose the resync period
The resync period allows wrapping an external, imperative API with a CompositeController, which partially addresses the "use-case for reconciling external state" described in #93. However, because there is a single resync period for the controller, if you want to minimize the latency of the status updates, then you might set a short resync period, resulting in unnecessary polling for resources that have settled.

This could be avoided by allowing the sync hook to say when it wants to be polled again for a given resource, eg by adding a `ResyncPeriodMilliseconds` field to `SyncHookResponse`. Then the hook could say that it wants to be polled frequently for a volatile resource, but infrequently once that resource has settled.

enisoc:
This adds a `resyncAfterSeconds` field (can be a decimal) to the `sync` hook response of CompositeController and DecoratorController. This field allows the hook response to request a resync of one particular object after a specified delay, for example if the hook knows it wants to recheck some external state for just that object. As a result, the controller-wide resyncPeriodSeconds can be kept at a
larger interval while being responsive when external state is more likely to change quickly for certain objects.

Closes #152

### DIFF
--- a/controller/composite/controller.go
+++ b/controller/composite/controller.go
@@ -238,6 +238,15 @@ func (pc *parentController) enqueueParentObject(obj interface{}) {
 	pc.queue.Add(key)
 }
 
+func (pc *parentController) enqueueParentObjectAfter(obj interface{}, delay time.Duration) {
+	key, err := common.KeyFunc(obj)
+	if err != nil {
+		utilruntime.HandleError(fmt.Errorf("couldn't get key for object %+v: %v", obj, err))
+		return
+	}
+	pc.queue.AddAfter(key, delay)
+}
+
 func (pc *parentController) updateParentObject(old, cur interface{}) {
 	// We used to ignore our own status updates, but we don't anymore.
 	// It's sometimes necessary for a hook to see its own status updates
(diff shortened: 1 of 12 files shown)


controller/composite/controller.go
```diff
@@ -489,7 +504,7 @@ func (pc *parentController) syncParentObject(parent *unstructured.Unstructured)
 
 	// Update parent status.
 	// We'll want to make sure this happens after manageChildren once we support observedGeneration.
-	if _, err := pc.updateParentStatus(parent, parentStatus); err != nil {
+	if _, err := pc.updateParentStatus(parent, syncResult.Status); err != nil {
```
crimsonfaith91: Will `syncResult.ParentStatus` be clearer here (to differentiate it is status of the parent, not child)?

crimsonfaith91:
/lgtm
Thanks!

controller/composite/controller.go
enisoc: After the change I'm making here, the field `.Status` is now coming from the canonical [SyncHookResponse](https://sourcegraph.com/github.com/GoogleCloudPlatform/metacontroller@29995294febd79eb4d29a4eeaacc6d8f01d55524/-/blob/controller/composite/hooks.go#L39) struct, so changing the name would mean changing the public API.

PR #160, opened 2019-03-31
Added Port name in Service-Per-Pod

shubhomoy:
Added port number as Port Name to allow creation of services having multiple ports


shubhomoy:
Hi!

I’ve signed the agreement as individual.

 

Thanks

 

Regards

Shubhomoy

